PR #1, opened 2016-10-02
Pass event object to handleDOMEvent handler

nirrek:
The signature in the docs for `handleDOMEvent` is:

```js
handleDOMEvent:: ?(view: EditorView, event: dom.Event) → bool
```

Currently an event object is not actually being passed as the second argument. Instead, the event name (string type) is being passed. This PR brings the implementation into agreement with the docs.


marijnh:
Thanks. Merged as 2dbe6bb and released as prosemirror-view 0.11.2

PR #2, opened 2016-10-24
Fix readHTML in Microsoft Edge

rafbm:
The `title` argument of `DOMImplementation#createHTMLDocument` is mandatory in Microsoft Edge.


marijnh:
Thanks for fixing that! Merged as 9baf9d5

PR #3, opened 2016-11-15
Fix focusing via editorView.focus() in Firefox

astashov:
It doesn't work currently because when you call editorView.focus(), it goes through `EditorView.focus() -> selectionToDOM()`, and there it tries to call `this.view.content.focus()`. `this` in that context is `Window`, and it doesn't have the `view` property. When I change it to just `view.content.focus()`, it seems like it fixes the issue in Firefox.


marijnh:
Thanks for fixing that! Merged as cc3df57

PR #4, opened 2016-11-18
Protect against global navigator and document

darobin:
The rest of the PM code base protects against these globals that will just blow up in Node, but the guards were missing here.

I've tried to stick to your coding style, LMK if it's not quite what you expect.

### DIFF
--- a/src/browser.js
+++ b/src/browser.js
@@ -1,10 +1,11 @@
-const ie_upto10 = /MSIE \d/.test(navigator.userAgent)
-const ie_11up = /Trident\/(?:[7-9]|\d{2,})\..*rv:(\d+)/.exec(navigator.userAgent)
+const has_navigator = typeof navigator != "undefined"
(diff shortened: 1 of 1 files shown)


marijnh:
You can't load the prosemirror-view module outside of the browser. It does nothing that makes any sense in a non-browser environment. Why are you trying to do that?

darobin:
ProseMirror is being used in a larger component, which is used in a yet larger set of stuff, the total bundle of which does get run through Node. The PM code will actually never get instantiated (because the editor is not mounted) but it will get compiled.

I could conditionally load `prosemirror-view` based on the environment but that gets ugly pretty quickly, and it interferes with bundlers.

marijnh:
It still seems wasteful and something of an invitation for trouble to load code that's written for the browser in node, but in this case it was indeed easy to avoid accessing `navigator` on load. I've pushed an alternative patch 50281c2 which is a little less likely to break again the next time someone edits that file.

PR #5, opened 2016-11-23
Remove delay before storing updated selection

rafbm:
We started using ProseMirror in a collaborative environment and found quite a major issue with cursor re-positioning when receiving changes from peers.

Here is a little video showing two windows, the latter which programatically types in text at a realistic pace. Pay attention to the uppermost window, where the only thing done is pressing the ➡️  key, first a couple times per second during a few seconds, then holding it non-stop for a few more seconds.

selection-before

You can see the cursor is often moved back to its previous position, and when the ➡️  key is held down it feels really buggish. The following explains the cause and solution to this problem.

Upon receiving a peer transform, ProseMirror applies the transform (which re-maps selection) based on the _last stored selection_. This is OK. The problem is _when_ the selection is actually stored into the state. Key and mouse events call the `SelectionReader#fastPoll` function instantaneously on each event, which then calls `startPolling`. However `startPolling` currently delays storing the selection by 50ms every time it is called.

This can be clearly illustrated by logging `selection`-type actions received by `onAction`:

```diff
 var view = window.view = new EditorView(document.querySelector(".full"), {
   state: state,
   onAction: function(action) {
+    if (action.type == "selection") {
+      console.log("selection change")
+    }
     view.updateState(view.state.applyAction(action))
   },
 })
```
selection-change-log

When I hold the ➡️  key, every call to `startPolling` clears the previous timeout and starts a new one with a 50ms delay, thus never storing the selection until I release the key for at least 50ms.

I changed the 50ms delay to 0ms, which ensures an updated selection is stored after every keystroke. This eliminates any cursor re-mapping issue during collaborative editing (if you see some lag, it’s a .gif lag):

selection-after

### DIFF
--- a/src/selection.js
+++ b/src/selection.js
@@ -31,7 +31,7 @@ class SelectionReader {
 
   startPolling(origin) {
     clearTimeout(this.polling)
-    this.polling = setTimeout(origin ? () => this.pollFunc(origin) : this.poller, 50)
+    this.polling = setTimeout(origin ? () => this.pollFunc(origin) : this.poller, 0)
   }
 
   fastPoll(origin) {


rafbm:
This is the code I used to have a window that programatically types text:

```html
<a href="javascript:" class="start-typing">Start typing</a>
-
<a href="javascript:" class="stop-typing">Stop typing</a>
```

```javascript
let typingTimeout
  , lastTypedCharacterIndex = 0
  , dummyText = 'Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Aenean commodo ligula eget dolor. Aenean massa. Cum sociis natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus. Donec quam felis, ultricies nec, pellentesque eu, pretium quis, sem. '

document.querySelector('.start-typing').onclick = () => {
  let type = () => {
    view.updateState(
      view.state.applyAction(
        view.state.tr.insertText(dummyText[lastTypedCharacterIndex]).action()
      )
    )
    lastTypedCharacterIndex += 1
    if (lastTypedCharacterIndex > dummyText.length - 1) {
      lastTypedCharacterIndex = 0
    }
    typingTimeout = setTimeout(type, 100)
  }
  type()
}

document.querySelector('.stop-typing').onclick = () => {
  clearTimeout(typingTimeout)
}
```

marijnh:
This is probably a good idea. I've merged it, if there's any fallout, we'll see later. Ideally we'd want to not cancel the timeout at all, to avoid having it pushed back by subsequent events, but arranging for that while preserving the distinction between fast and slow polling would be somewhat more cumbersome, and if this simple change does the trick I prefer it.

PR #6, opened 2017-02-10
Allow place function to return view content

kapouer:
This is useful in case the content just need to be an existing node.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -34,7 +34,7 @@ class EditorView {
     this.content = document.createElement("div")
 
     if (place && place.appendChild) place.appendChild(this.content)
-    else if (place) place(this.content)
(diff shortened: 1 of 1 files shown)


marijnh:
I think the concept or turning any node into a ProseMirror editor is sound but I don't find the way it works here very clean. Unfortunately, I'm having trouble coming up with something better, so I'm leaving this open for a bit in the hope of finding a good solution.

marijnh:
I ended up going with a slightly different interface in [this patch](https://github.com/ProseMirror/prosemirror-view/commit/6c43f25a1177d9ba3d6ad2f06acacb00292b3762) (you pass `{mount: myDOMNode}` as place).

kapouer:
Adopted & works as expected. Thank you.

PR #7, opened 2017-03-20
Ensure selection is definitely collapsed - fixes Blink Shadow DOM bug

bedeoverend:
This checks that a selection's ranges are all collapsed instead of just relying on the `isCollapsed` property. This can occur in both v0 and v1 Shadow DOM implementations in Chrome - see [this Blink bug](https://bugs.chromium.org/p/chromium/issues/detail?id=447523).

Fixes [this issue](https://github.com/ProseMirror/prosemirror/issues/588)

### DIFF
--- a/src/domchange.js
+++ b/src/domchange.js
@@ -3,7 +3,7 @@ const {Selection} = require("prosemirror-state")
 const {Mapping} = require("prosemirror-transform")
 
 const {TrackMappings} = require("./trackmappings")
-const {selectionBetween} = require("./selection")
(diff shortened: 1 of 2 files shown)


marijnh:
Thanks! Merged and followed up with 1d6df3e (since browser don't actually support multiple selections, checking range 0 is enough).

bedeoverend:
@marijnh yeah I didn't quite know what the deal was with multiple ranges - thanks for clearing that up :)

PR #8, opened 2017-04-05
Fix typo in dragstart dataTranser -> dataTransfer

kapouer:


marijnh:
Thanks for spotting that!

PR #9, opened 2017-04-10
Check for rangeCount and anchorNode in selection check

bedeoverend:
Adds a check in `hasFocusAndSelection` to check if the selection has an `anchorNode` as well as `rangeCount` > 0

This fixes https://github.com/ProseMirror/prosemirror/issues/614 where Shadow DOM in Chrome 57 incorrectly reports the selection as still having a range even when there's no selection and therefore no `anchorNode`.

### DIFF
--- a/src/selection.js
+++ b/src/selection.js
@@ -282,5 +282,5 @@ exports.selectionBetween = selectionBetween
 function hasFocusAndSelection(view) {
   if (view.editable && view.root.activeElement != view.dom) return false
   let sel = view.root.getSelection()
-  return sel.rangeCount && view.dom.contains(sel.anchorNode.nodeType == 3 ? sel.anchorNode.parentNode : sel.anchorNode)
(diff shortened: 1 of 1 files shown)


bedeoverend:
@marijnh would just changing the line to check for _only_ `anchorNode` have the same affect? i.e.

```js
function hasFocusAndSelection(view) {
  if (view.editable && view.root.activeElement != view.dom) return false
  let sel = view.root.getSelection()
  return sel.anchorNode && view.dom.contains(sel.anchorNode.nodeType == 3 ? sel.anchorNode.parentNode : sel.anchorNode)
}
```

Thanks

marijnh:
Yeah, using only `anchorNode` should be fine -- if there's a range but no anchornode, that's exactly the bug we are trying to handle.

bedeoverend:
Cool, I'll update the PR with just checks for `anchorNode`

bedeoverend:
Done. Is there anything else you need on this PR?

marijnh:
Nope. Merged as 3edbceb

bedeoverend:
Awesome - thanks for the quick response @marijnh

PR #10, opened 2017-04-12
Add chrome browser check

bedeoverend:
Adds in a `chrome` property to the browsers file. This allows the selection bug [check](https://github.com/ProseMirror/prosemirror-view/blob/master/src/dom.js#L60) which was added to fix [Shadow DOM failing in chrome](https://github.com/ProseMirror/prosemirror/issues/588)


bedeoverend:
Not sure about the User-Agent string check on this. Just an assumption that `/Chrome/` would be enough

marijnh:
Oops, I pushed 931e286 before I saw this. That one also makes sure there's a slash after `Chrome` to avoid false positives for user agent strings that randomly throw the word `Chrome` in there.

PR #12, opened 2017-06-14
Fix errors when view is inside an iframe

jacwright:
This fixes issues with prosemirror view inside an iframe. Using iframes can provide a performance boost by sandboxing the layout inside and outside the iframe. Most of prosemirror already works inside iframes.

### DIFF
--- a/src/domcoords.js
+++ b/src/domcoords.js
@@ -41,7 +41,7 @@ exports.scrollRectIntoView = scrollRectIntoView
 // when the size of the content above changes.
 function storeScrollPos(view) {
   let rect = view.dom.getBoundingClientRect(), startY = Math.max(0, rect.top)
-  let doc = view.dom.defaultView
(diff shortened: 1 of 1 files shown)


src/domcoords.js
```diff
@@ -40,6 +41,7 @@ exports.scrollRectIntoView = scrollRectIntoView
 // when the size of the content above changes.
 function storeScrollPos(view) {
   let rect = view.dom.getBoundingClientRect(), startY = Math.max(0, rect.top)
+  let doc = view.dom.defaultView
```
marijnh: This looks like it'll yield `undefined`, and definitely not the document object (since that would be `ownerDocument`). Am I missing something?

jacwright: Oh that was bad copy/paste. :(

marijnh:
Thanks for fixing that. Merged as b3be7d7

PR #13, opened 2017-06-17
Additional iframe fixes

jacwright:
These were not throwing errors that I could see, but were breaking
transaction.setSelection commands.

### DIFF
--- a/src/input.js
+++ b/src/input.js
@@ -359,15 +359,16 @@ editHandlers.input = view => DOMChange.start(view)
 function captureCopy(view, dom) {
   // The extra wrapper is somehow necessary on IE/Edge to prevent the
   // content from being mangled when it is put onto the clipboard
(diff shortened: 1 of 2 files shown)


src/input.js
```diff
@@ -359,15 +359,16 @@ editHandlers.input = view => DOMChange.start(view)
 function captureCopy(view, dom) {
   // The extra wrapper is somehow necessary on IE/Edge to prevent the
   // content from being mangled when it is put onto the clipboard
-  let wrap = document.body.appendChild(document.createElement("div"))
+  let doc = view.root
```
marijnh: `view.root` may be a shadow root, which has neither `body` nor `createElement` methods. Is there any problem with using `ownerDocument` here as well?

jacwright: No, I didn't realize root could be a shadow root. Will update.

marijnh:
Thanks. Merged as bae34c6ad93

PR #14, opened 2017-06-17
Allow the DOMChange timeout to be customized

jacwright:
This 20ms timeout is to fix issues with Android Chrome, but it makes a
jitter in the editor when plugins cancel/modify user input.

This change allows the implementor to change the timeout. This may be
done statically if Android is not a supported platform, or the
implementor may use means to detect the platform and change the timeout
based on that.

This may be a stop-gap until someone can find time to do some hard testing on Android for a better solution.

### DIFF
--- a/src/domchange.js
+++ b/src/domchange.js
@@ -12,7 +12,7 @@ class DOMChange {
     this.state = view.state
     this.composing = composing
     this.from = this.to = null
-    this.timeout = composing ? null : setTimeout(() => this.finish(), 20)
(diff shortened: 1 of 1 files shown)


marijnh:
🤢 Eww. But okay, as long as it's understood that this isn't part of the official interface and _will_ be dropped in the future, I'll let it in for now.

jacwright:
Exactly, and nice part is, when it is dropped, it won't be breaking

PR #15, opened 2017-06-20
Add coordsAtRange

jacwright:
Provide helper to get coordinates for the given range. This unlocks
floating menu plugins to display over selected text.

### DIFF
--- a/src/domcoords.js
+++ b/src/domcoords.js
@@ -238,6 +238,21 @@ function coordsAtPos(view, pos) {
 }
 exports.coordsAtPos = coordsAtPos
 
+// : (EditorView, number, number) → {left: number, top: number, right: number, bottom: number}
+// Given a range in the document model, get a bounding box of the
+// entire range, relative to the window.
+function coordsAtRange(view, from, to) {
+  if (from === to) return coordsAtPos(view, from)
+  let {node: fromNode, offset: fromOffset} = view.docView.domFromPos(from)
+  let {node: toNode, offset: toOffset} = view.docView.domFromPos(to)
+  let range = document.createRange()
+  range.setStart(fromNode, fromOffset)
+  range.setEnd(toNode, toOffset)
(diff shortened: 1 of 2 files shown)


marijnh:
For positioning relative to the selection, I've had good results from just directly using the DOM selection to get the DOM positions. Have you tried that? I'd prefer a solution that doesn't involve adding stuff to our interface.

jacwright:
That works for selection since there is a DOM equivalent to the ProseMirror selection object. It doesn't work for getting the coordinates of a range that isn't selected. This is necessary for external tooltips on portions of the document that aren't selected (necessary when the content is inside a scrollable area), implementing a custom Find/Find Next feature to scroll down to the next instance of a word (though you could arguably use selection for this), showing the position of certain items in a generated minimap view of long content, etc.

I agree with trying to keep the API surface small. There just isn't a public way to get the DOM range for a given ProseMirror range. Is this simply because `docVew` and `domFromPos` isn't documented yet?

Here are some other options to consider:

* Make a general `rangeToDom(from, to)` method on EditorView that returns a dom range (not sure I like this)
* Update `coordsAtPos(pos)` to `coordsAtPos(pos, [to])` making it do double duty, keeping the method count the same while opening up this capability. Could rename it to reflect, though that becomes a breaking change, so is less ideal.
* Document `view.docView.domFromPos` and `view.inDOMChange.mapping.invert().map(pos)` and add this workflow to a cookbook for those wanting to achieve this and not knowing how.

I'd favor the PR as it currently exists, or the second alternative above. Since `coordsAtPos` already exists, implementors looking to get coords for a range will be looking at that method or something similarly named.

marijnh:
Attached patch implements your third proposal. I decided against merging this as-is because I can see people needing `getClientRects` in some cases, or other obscure information about the DOM, and it seems more future-proof to provide the primitive rather than a complicated high-level interface.

jacwright:
That works. Thank you!

PR #16, opened 2017-06-20
Fixes bug introduced in #13

jacwright:
This bug manifests when pressing Esc inside a block to select the block.


marijnh:
Thanks, merged

PR #17, opened 2017-07-07
Fix wrong node match during sync

torifat:
Consider the following document:
```
Content 1
[selection]
[Very big node with lots of nested nodes]
x
```
If I type `x` in place of `[selection]`. Then during `findNodeMatch` it will match the new node with the last `x` (since `content` & `marks` same) and remove `[Very big node with lots of nested nodes]` and recreate it afterward.

Same thing happens for new lines too. Though the end result is same. But for very big documents it has performance implications.

### DIFF
--- a/src/viewdesc.js
+++ b/src/viewdesc.js
@@ -542,7 +542,7 @@ class NodeViewDesc extends ViewDesc {
   }
 
   matchesNode(node, outerDeco, innerDeco) {
-    return this.dirty == NOT_DIRTY && node.eq(this.node) &&
+    return this.dirty == NOT_DIRTY && node == this.node &&
       sameOuterDeco(outerDeco, this.outerDeco) && innerDeco.eq(this.innerDeco)
   }
 


marijnh:
Node object equality is also not something you can rely on -- a node is allowed to appear in a document multiple times, and it is also not unlikely for a transaction to end up replacing a node with a new node that has the same content. So the choice for deep comparison is intentional here -- with `==` you get other issues. It would, for example, completely kill performance in a setup that serializes and de-serializes the document during editing, which is a perfectly valid thing to do. In general, this data has structure, not identity.

That being said, it would be nice to find a heuristic that avoids the extra work in situations like this. The obvious approach would be to see if the skipped nodes occur somewhere ahead, and refuse to skip them then, but that's a little complicated by the fact that the nodes are iterated over using `iterDeco`, which may split text nodes. Still, text nodes aren't the problem (they are cheap to re-render anyway), so looking ahead in the plain nodes would probably be okay.

torifat:
> The obvious approach would be to see if the skipped nodes occur somewhere ahead, and refuse to skip them

If we look ahead and refuse to skip them then we end up not skipping anything.
Sorry, I think I didn't get it properly.

> Still, text nodes aren't the problem (they are cheap to re-render anyway), so looking ahead in the plain nodes would probably be okay.

I tried only destroying the node (`this.destroyBetween(this.index, i)`) if it's a text node. It fixes the problem for insertion but now we are having the same problem when removing a node.

marijnh:
I think patch 139cb0d7f3 should solve the issue you're addressing here.

torifat:
Thanks :)

PR #18, opened 2017-07-11
Fix explosion on creating editor when editable returns false.

alexf101:
The SelectionReader is created in the EditorView constructor before calling 'initInput'.

If the view isn't editable, this starts a poller to keep the selection in sync with the DOM, which ends up calling `this.view.domObserver.flush()`.

Since the `domObserver` gets created by initInput, which hasn't been called yet, this doesn't exist and explodes.

I'm not quite sure why my usage is triggering this bug, I assume other people's isn't? I'm rendering this editor output on a tooltip which is getting moved about, so perhaps it's something to do with that - regardless, this seems like a clear bug to me. Creating the `domObserver` before the `SelectionReader` appears to fix it.

Note that the SelectionReader can also generate a transaction before init returns, which I'm not sure if is considered a problem in ProseMirror.

Alternatively it may be better to disable the selection api while the editor isn't editable?

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -45,8 +45,9 @@ class EditorView {
     this.docView = docViewDesc(this.state.doc, computeDocDeco(this), viewDecorations(this), this.dom, this)
 
     this.lastSelectedViewDesc = null
+    initInput(this) // Must be done before creating a SelectionReader
+
     this.selectionReader = new SelectionReader(this)
-    initInput(this)
 
     this.pluginViews = []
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks, I've merged this as 02a04548bb5c2d and added 5dbf517628f4ffab1 to address the firing of a transaction before the editor has finished initializing.

It may well be that no one has been working with editors that start out uneditable -- the library is still young.

PR #19, opened 2017-07-18
Make Decoration.from and Decoration.to public API

bradleyayers:
I propose making `from` an `to` public API, to open up some new possibilities with decorations.

My use-case is having decorations that can be either 'active' or 'inactive'. An example is inline comments use inline decorations to highlight text, and use different colors depending on whether or not that specific comment is being viewed (e.g. in the document's sidebar).

Given a decoration, it's not currently possible to create a new decoration over the same range (since `from` and `to` are not exposed).

### DIFF
--- a/src/decoration.js
+++ b/src/decoration.js
@@ -91,7 +91,9 @@ class NodeType {
 // static members of this class for details.
 class Decoration {
   constructor(from, to, type) {
+    // :: number
     this.from = from
+    // :: number
     this.to = to
     this.type = type
   }


wes-r:
What about finding the existing decoration (https://prosemirror.net/docs/ref/#view.DecorationSet.find) and then changing the attributes on it? (haven't done this, so I don't know if it'll work).

bradleyayers:
I don't see how that'll work, as decorations don't expose any builder/mutation methods to change attributes. It's definitely an option I'm open to though — it's the same spirit as `Transaction` builder chaining or any immutable builder pattern really 👍 .

marijnh:
Agreed, it makes sense for these to be public. Merged.

PR #20, opened 2017-07-19
Clarify the coordinate space used by coordsAtPos

bradleyayers:
I didn't understand the documentation when it referred to "screen coordinations", so I edited it to use "viewport" which is the term MDN uses.


marijnh:
Thanks!

PR #21, opened 2017-08-02
Fix wrapMap content order

MengWeiChen:
When I copy the **tr** node in the editor, then it give me` <tbody><table><tr>....</tr></table></tbody>  `

I find out serializeForClipboard function will wrap the wrapMap content **sequentially** if the nodeName is in the wrapMap.

For example, the "tr" key map to wapMap is `['table', 'tbody']` , so serializeForClipboard will wrap the "tr" node around the "table" node first, then wrap around the "tbody" node.

So I reverse the wrapMap content to fix this issue.


marijnh:
I've pushed an alternative patch (931f867) because `wrapMap` is  being used by `readHTML` as well, which _does_ expect it to be in outer-to-inner order.

PR #23, opened 2017-10-17
Show text cursor over text in list items

adrianheine:
FIX: Show text cursor, not pointer, over text in list items.

### DIFF
--- a/style/prosemirror.css
+++ b/style/prosemirror.css
@@ -13,6 +13,10 @@
   cursor: default;
 }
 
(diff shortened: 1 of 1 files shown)


marijnh:
I think I'd rather just drop the `cursor` override altogether and leave the default behavior, now that we're not adding `pointer-events` styles to lists anymore. See be7905c. Does that address the problem you had?

adrianheine:
I prefer your solution, thanks.

PR #24, opened 2017-10-22
Update comment on selectionchange event in FF

adrianheine:
Firefox supports this event since 52.


marijnh:
Nice — the polling fallback isn't great. Would be nice if we can eventually drop it.

PR #25, opened 2017-11-16
Fix arrowing left into a node

cmcelwain822:
When selecting horizontally to the left, we shouldn't find `descAt($head.pos)` since that won't correspond to the node we are trying to select.


marijnh:
Thanks! But you're computing `nodePos` and accessing `node.nodeSize` at a point where `node` might be null. Could you revise the code to move that expression into the call to `descAt` (which is only made when `node` isn't falsy)?

cmcelwain822:
Good point! Will do.

marijnh:
Looks great! Merged as 0353426

PR #26, opened 2017-11-22
Fix wrong step on backspace for ambiguous content

torifat:
Closes ProseMirror/prosemirror#726

### DIFF
--- a/src/domchange.js
+++ b/src/domchange.js
@@ -211,7 +211,9 @@ function readDOMChange(view, mapping, oldState, range, allowTypeOver) {
   let parse = parseBetween(view, oldState, range)
 
   let doc = oldState.doc, compare = doc.slice(parse.from, parse.to)
-  let change = findDiff(compare.content, parse.doc.content, parse.from, oldState.selection.from)
+  // Shift preferedStart by 1 when Backspace is pressed
+  let preferedStart = view.lastKeyCode === 8 ? oldState.selection.from - 1 : oldState.selection.from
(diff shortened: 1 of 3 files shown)


marijnh:
Looks great! Can you add a test? (I think what would work well is to add it to `test/test-domchange.js` and have a test that does two deletions like this, one with `lastKeyCode` set to 8, and one with it set to something else.)

Also, it may be worth resetting `lastKeyCode` after each change, or keeping a time counter with it and ignoring it if it's too old, to avoid spurious effects when a non-key-related change happens after a backspace press.

torifat:
Yes, I can. I actually tried to add a test to `test/test-domchange.js` but didn't because the output is same. I didn't think about setting `lastKeyCode` manually though. Thanks for the suggestion.

I was also worried about unwanted states. So, I manually tried <kbd>Shit + Backspace</kbd>, <kbd>Ctrl + Backspace</kbd>, etc. Keeping a time counter is a great idea, it didn't occur to me.

torifat:
I tried to incorporate your suggested changes. Let me know if you have any opinion about them 😃

marijnh:
Thanks for following up! I've merged your work as eb194674f5095f and followed up with 0398b2b4076330de to make it work for ambiguous ctrl-backspace. The changes have been released as prosemirror-view 1.0.3

torifat:
Thanks for the release. Don't know why didn't I think about `dispatchTransaction` 🤦‍♂️

PR #27, opened 2017-11-25
Expand documentation for `handle*ClickOn`

adrianheine:
See ProseMirror/prosemirror#525. @marijnh Is that something you would accept in the reference manual? @styts would that have answered your question?


marijnh:
I'm not a fan of repetitive, formulaic docs—I'd prefer either a reference like 'see handleClickOn for further notes` or a more succinct formulation in the second two instances.

PR #28, opened 2017-11-27
Ensure posAtCoords returned non-null result before accessing pos

mjmasn:
Fixes ProseMirror/prosemirror#728

Linter and tests are passing :)


marijnh:
Thank you! Merged as a309087

PR #29, opened 2017-12-20
Be less strict to find out if view has focus

kapouer:
This allows client code to workaround https://github.com/ProseMirror/prosemirror/issues/745

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -188,7 +188,7 @@ export class EditorView {
   // :: () → bool
   // Query whether the view has focus.
   hasFocus() {
-    return this.root.activeElement == this.dom
+    return this.dom.contains(this.root.activeElement)
   }
 
   // :: ()
--- a/src/selection.js
+++ b/src/selection.js
@@ -287,7 +287,7 @@ export function selectionBetween(view, $anchor, $head, bias) {
 }
 
 function hasFocusAndSelection(view) {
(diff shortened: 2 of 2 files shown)


kapouer:
I suppose this cannot be accepted without stronger arguments.
The only reason i made the PR is to leave some clues about how i worked around the issue.

kapouer:
Oh i just found a better approach here.

src/index.js
```diff
@@ -188,7 +188,7 @@ export class EditorView {
   // :: () → bool
   // Query whether the view has focus.
   hasFocus() {
-    return this.root.activeElement == this.dom
+    return this.dom.contains(this.root.activeElement)
```
marijnh: This will also return true if the focus is in a completely unrelated descendant element (for example a textarea in a node view). In general, the conditions under which something should be considered part of the same editor are fuzzy, and I'd rather stick to a simple model than introduce confusion here.

kapouer: I understand your position, i'm just sharing ideas. For now i'm happy with my "crazy" patch: it keeps view focused even if going into an island inside view; and it's meant to be used with https://github.com/pageboard/pagecut/blob/1017f8c9263414de17428ea6c2614a1574b0fbbf/src/specs.js#L388-L396.

kapouer:
Not useful to keep around, closing.

PR #30, opened 2017-12-22
Allow markview to specify a contentDOM element

dxparker:
Use case: Use a markview to render hyperlink as a preview of the content
it links to for certain content types, e.g. a video thumbnail.
Otherwise allow prosemirror to manage hyperlinked text using contentDOM.

### DIFF
--- a/src/viewdesc.js
+++ b/src/viewdesc.js
@@ -464,7 +464,9 @@ class MarkViewDesc extends ViewDesc {
     let custom = customNodeViews(view)[mark.type.name]
     let spec = custom && custom(mark, view)
     let dom = spec && spec.dom || DOMSerializer.renderSpec(document, mark.type.spec.toDOM(mark, inline)).dom
-    return new MarkViewDesc(parent, mark, dom)
+    let markDesc = new MarkViewDesc(parent, mark, dom)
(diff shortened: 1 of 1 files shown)


dxparker:
Hi @marijnh, any thoughts on this PR? The intent is to allow hyperlinks to either render as traditional marks where the content they apply to is managed by prosemirror, or render as opaque elements using a custom view depending on the href of the hyperlink.

torifat:
Why not use [Decorations](http://prosemirror.net/docs/ref/#view.Decorations) for this? IMO it doesn't make sense to add `contentDOM` in `mark`.

dxparker:
Hi @torifat, I started a [thread on the forum](https://discuss.prosemirror.net/t/unfurling-links/1191) to describe the situation in more detail and solicit suggestions. I don't see how decorations could solve the problem. Can you elaborate there?

marijnh:
I've merged d6b25fb0cb2fc74a20a5bd2299011d4cb2734d59 instead, which does the same thing slightly cleaner (adding an argument to the `MarkViewSpec` constructor instead of mutating the object after construction).

dxparker:
Thank you Marijnh. I agree, the constructor argument is preferable.

PR #31, opened 2018-02-28
correcting a typo

saranrapjs:
A luck of build timing for our app led us to run into this just recently — I promise we're not just watching every commit! :)


marijnh:
Oh, that was a bad mistake (I guess I do need to set up tests for UI stuff like this somehow). Merged and released as 1.1.1!

PR #32, opened 2018-05-01
Check that `document` is defined

mjmeintjes:
Check that document is defined before determining browser capabilities. 

Otherwise this fails inside a web worker context.


marijnh:
Thanks. Merged as f1be5eb

PR #33, opened 2018-06-30
support for nodes those not under document.body

enginaryum:
Hey guys,

when EditorView.updateState called, scrollRectIntoView() raising following error:
image

As I've seen the issue is caused that our editor is not under document.body. 
image

A quick fix is to break scroll loop if parent is document itself. 

We are excited to use this editor, can you please add support for such a case.

### DIFF
--- a/src/domcoords.js
+++ b/src/domcoords.js
@@ -10,7 +10,7 @@ export function scrollRectIntoView(view, rect) {
   let doc = view.dom.ownerDocument, win = doc.defaultView
   if (scrollMargin == null) scrollMargin = 5
   for (let parent = view.dom;; parent = parentNode(parent)) {
-    if (!parent) break
+    if (!parent || parent === document) break
     let atBody = parent == doc.body
     let bounding = atBody ? windowRect(win) : parent.getBoundingClientRect()
     let moveX = 0, moveY = 0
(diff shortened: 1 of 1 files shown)


kapouer:
Sorry to bother you about that detail, but can you tell how different browsers behave with this DOM node outside of body ? I thought it was not possible to do that...

enginaryum:
I've tested this in Firefox, Chrome, Opera and Safari. All seems working fine, not sure about other browsers. Couldn't have a chance to test on them. 

We are using this structure on our Chrome extension and it's working fine on Chrome. So honestly, we are not worried much about other browsers.

You can test by appending an element to html (document.documentElement) as follows:

```
var node = document.createElement('div')
document.documentElement.appendChild(node)
```

marijnh:
Well, I also did not know you could do that.

Could you check for `parent.nodeType != 1` instead of `== document`, to cover any other weird situations?

enginaryum:
@marijnh that makes sense, updated!

kapouer:
@enginaryum this is too good to be true ! It opens up so many possibilities... there must be something wrong with it :)

marijnh:
Oh, I just realized this will fail to scroll the window, because it breaks out of the loop without `atBody` ever being true. I've pushed an alternative patch, e1c53c4, which avoids that.

kapouer:
For the record, one can add and render a dom node child of html, besides body.
The html parser might not like it, but using `appendChild` works all down to IE10 (but rendering that node does happen only in IE11 and above). Did not try safari.

PR #34, opened 2018-07-26
Bind this to view in dispatchTransaction prop

RichieAHB:
A suggested fix for [this issue](https://github.com/ProseMirror/prosemirror/issues/826). Happy for this to be closed if there is an alternative solution and this just encourages bad practices!

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -330,7 +330,7 @@ export class EditorView {
   // easily passed around.
   dispatch(tr) {
     let dispatchTransaction = this._props.dispatchTransaction
-    if (dispatchTransaction) dispatchTransaction(tr)
+    if (dispatchTransaction) dispatchTransaction.call(this, tr)
     else this.updateState(this.state.apply(tr))
   }
 }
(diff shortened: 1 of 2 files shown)


marijnh:
Thanks. Merged as eca97f85123963 and released as 1.4.0

RichieAHB:
@marijnh I completely forgot that this change actually broke two tests in `selection-test`. However, they pass when run on their own so they seem to be related to `tempEditor` but I didn't have time to investigate. Happy to look when I have time but thought you may have a better idea of what was going on.

marijnh:
I'm working on a 2010-era Thinkpad x201, so I doubt my laptop is faster than yours, but maybe some other factor interfered. Which test, precisely, failed for you? And are you sure it wasn't related to the focus not being in the document when the tests ran?

RichieAHB:
I was the tests that contain `posAtCoords` - they run intermittently for
me. I can’t imagine it was to do with focus as I refreshed in the same
manner each time. It may well be that I was running other things while
running the tests too. Either way, after checking it the test failed for me
before this commit, if I upped the number in setTimeout it passed every
time so I assume it’s just the magic number in setTimeout.

PR #35, opened 2018-07-28
Add ability to run any kind of transaction during init

RichieAHB:
With my last PR it allowed me to run some transactions (specifically regarding node views). This should add the ability to run any transactions that can be run against an unfocussed view by just queuing them until init is complete.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -34,6 +34,9 @@ export class EditorView {
     this._root = null
     this.focused = false
 
+    this.initialized = false;
+    this.transactionQueue = [];
+
     // :: dom.Element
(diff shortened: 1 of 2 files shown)


marijnh:
Seeing how you had to do something completely terrible (run a transaction from a node serialization function) to test this, I'm not really convinced this is a useful feature. If it _is_ really needed, it can be implemented in your `dispatchTransaction` callback. I really, really don't want this kind of side-effects-gone-wild style feature in the core library.

RichieAHB:
Ok no problem, I thought you might say that but some time’s it’s easier to
make a PR to explain an idea!

PR #36, opened 2018-09-26
Add a few enhancements to scroll behaviors

wes-r:
1. Allow `scrollThreshold` and `scrollMargin` to be side specific (top,
right, bottom, left).
2. Start scrolling from position of `selection.head` AND NOT `view.dom`. This
way nodes that have internal scrolling (usually horizontal) can get
scrolled (useful for tables).
3. Try to scroll entire selection into view. This way tabbing within a table, which highlights the cell contents, will scroll the entire cell instead of just one side.

I needed 1 in order to differentiate top/bottom vs left/right AND I've got an absolute positioned toolbar that sits above the top part of my editor, thus I need a `scrollThreshold.top` to be greater than `scrollThreshold.bottom`.

Here's what 2 and 3 enable:
sep-25-2018 23-52-13

Happy to make any changes requested, would love to see this merged in to prevent necessitating the use of my own fork :)

### DIFF
--- a/src/domcoords.js
+++ b/src/domcoords.js
@@ -6,22 +6,27 @@ function windowRect(win) {
 }
 
 export function scrollRectIntoView(view, rect) {
-  let scrollThreshold = view.someProp("scrollThreshold") || 0, scrollMargin = view.someProp("scrollMargin")
+  let scrollThreshold = view.someProp("scrollThreshold") || 0, scrollMargin = view.someProp("scrollMargin") || 5
+  if (Number.isInteger(scrollThreshold))
+    scrollThreshold = {top: scrollThreshold, right: scrollThreshold, bottom: scrollThreshold, left: scrollThreshold}
+  if (Number.isInteger(scrollMargin))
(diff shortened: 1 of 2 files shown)


marijnh:
I'm not interested in 3 — even if you fix it to no longer assume `from` is always to the left of to `to` (which it isn't in RTL text or across lines), it no longer guarantees that the head will be scrolled into view, which is a main requirement.

1 and 2 sound reasonable. You can't rely on `Number.isInteger` since we target ES5 and that's ES6. The integer part seems irrelevant anyway, since `typeof == "number"` is enough. Could you do the dispatch in an accessor (function getSide(value, side) { return typeof value == "number" ? value : value[side] }`) instead of always allocating the objects?

wes-r:
Cleaned up 1 & 2 as you described.

As for 3, I've removed it. There may be a way to switch it to `head`/`anchor` and signal prioritization of scrolling `head` but it is probably overly complicated for such an edge case, especially considering it won't work right for multi-line selections.

marijnh:
Merged and followed up with 2742295

PR #37, opened 2018-10-08
Remove SelectionReader focus and blur listeners

kapouer:
Came across this when i was destroying an editor and recreating it on the same dom node.


marijnh:
Thanks for spotting that! Merged and released as 1.6.2

PR #39, opened 2018-11-15
doc: typo in decoration

kapouer:
I like positons ! I'm pretty sure there is some kind of deep duality between states, transactions and matter, energy (or fermions, bosons). Anyway it's almost certain you did not had that in mind here ;)


marijnh:
Thanks!

PR #40, opened 2018-11-15
Let docViewDesc be a customNodeViewDesc

kapouer:
Please squash the commits i had some indentation issue.

### DIFF
--- a/src/viewdesc.js
+++ b/src/viewdesc.js
@@ -665,7 +665,7 @@ class NodeViewDesc extends ViewDesc {
 // and used by the view class.
 export function docViewDesc(doc, outerDeco, innerDeco, dom, view) {
   let custom = view.nodeViews[doc.type.name]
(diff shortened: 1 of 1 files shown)


marijnh:
This doesn't really seem to follow the contract for custom node views—the dom node has already been allocated, whereas custom views get to create their own DOM. Also, the doc isn't at position 0, its _content_ starts at position 0. In general, the top node plays a different role than its content nodes, and trying to treat it the same is problematic.

kapouer:
Okay ! surely i did not figure out how to make prosemirror properly ignore when specific dom nodes are added. My first try was with `ignoreMutation` because every block node has a custom nodeView in my setup - and this fails when it comes to childNodes of the root dom node.

marijnh:
> and this fails when it comes to childNodes of the root dom node.

What kind of child nodes are these?

kapouer:
Typically a div node added by https://github.com/wilddeer/stickyfill.

PR #41, opened 2018-12-17
Use nearestDesc to find the real draggable target

rev-hikari:


marijnh:
What does this solve? The `target` of a drag event will be the element that is being dragged (in the browser's view). Dragging one of its ancestors might not be what the user intends.

rev-hikari:
The `event.target` is from a `mousedown` event, not a `drag*` event. That means, `event.target` may be a descendant of the expected draggable node.

screenshot

Here's the example: https://codesandbox.io/s/jprm772y25

In the example, try to drag each part ("A" / "B") of the block will cause `draggable` attribute added on the descendant of the block, and only that descendant will be shown under the pointer while dragging.

screenshot

PR #42, opened 2018-12-31
Remove unnecessary code

rev-hikari:


marijnh:
This code provides a way for people to target CSS to a focused editor, and as such is _not_ unnecessary.

rev-hikari:
Adding and removing class `ProseMirror-focused` is done in [input.js](https://github.com/ProseMirror/prosemirror-view/blob/33fe4a8/src/input.js#L542-L554), just like what `ProseMirror-hideselection` does (in [selection.js](https://github.com/ProseMirror/prosemirror-view/blob/33fe4a8/src/selection.js#L211-L216)).

Such logic in `computeDocDeco` is unnecessary, and may have some performance overhead (more frequent diffs in [viewdesc.js](https://github.com/ProseMirror/prosemirror-view/blob/33fe4a8/src/viewdesc.js#L857-L864)).

rev-hikari:
Another problem is, `computeDocDeco` will not be called each time `view.focused` changed. That means, checking `view.focused` here is unreliable, and developers can't use `props.attributes` to customize focused class on editor.

marijnh:
I believe that's why the code you removed is there. Can you come up with a situation where the focused class will go out of sync with the editor's focus status with the current code?

rev-hikari:
`computeDocDeco` will be called each time `view.state` changed, but focus / blur the editor (without modify selection) doesn't make any changes on the editor state.

marijnh:
Right, that's why there are separate class-toggling calls in the focus/blur handlers.

rev-hikari:
No, it doesn't overwrite existing class names, `class` attribute is processed separately in [`patchAttributes`](https://github.com/ProseMirror/prosemirror-view/blob/33fe4a8/src/viewdesc.js#L857-L864). It will find the difference between current and previous class attribute returned by the decoration, and apply it to the dom.

marijnh:
I see. I've merged this as da4b311

PR #43, opened 2019-01-07
Control wrapping behaviour for block marks

eshvedai:
Our discussion on this: https://discuss.prosemirror.net/t/block-marks-wrapping-behaviour/1718

Let me know if it looks good and I'll add some tests.

### DIFF
--- a/src/viewdesc.js
+++ b/src/viewdesc.js
@@ -471,10 +471,9 @@ class CursorWrapperDesc extends WidgetViewDesc {
 // necessary.
 class MarkViewDesc extends ViewDesc {
   // : (ViewDesc, Mark, dom.Node)
-  constructor(parent, mark, dom, contentDOM, disableWrapping) {
+  constructor(parent, mark, dom, contentDOM) {
     super(parent, [], dom, contentDOM)
     this.mark = mark
-    this.disableWrapping = disableWrapping
   }
 
   static create(parent, mark, inline, view) {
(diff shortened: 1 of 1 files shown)


marijnh:
This approach will also prevent reusing of mark nodes during redraw—you'll have to find some way to do this that doesn't have that as a side effect (maybe do the check further up, in `syncToMarks`, by always exiting non-spanning marks when they already have a child node).

Also, the doc comments for `MarkSpec` in prosemirror-model should be updated to mention this property.

eshvedai:
you're right, it is pretty bad. I updated the PR, looks like it doesn't re-create nodes with this change, have a look!

marijnh:
Yes, that looks a lot better. But this will also need an accompanying PR in prosemirror-model that documents the spec property and implements it in `DOMSerializer`.

eshvedai:
Sure thing. Just to confirm my understanding, what do you mean exactly by implementing it in `DOMSerializer`? What's the expected result?

marijnh:
> what do you mean exactly by implementing it in DOMSerializer? What's the expected result?

When serializing adjacent nodes with `spanning: false` marks across them, the marks should be rendered separately per node, just like in the editor view.

eshvedai:
Finally got the time to implement it: https://github.com/ProseMirror/prosemirror-model/pull/32
Let me know if I got it right.

marijnh:
Thanks! Merged as  25eb2af840a71 and released as 1.7.0.

PR #44, opened 2019-01-31
Handle Safari’s keydown on compositionend

kevinzang:
Implements patch for https://github.com/ProseMirror/prosemirror/issues/880.

### DIFF
--- a/src/domchange.js
+++ b/src/domchange.js
@@ -102,11 +102,9 @@ export class DOMChange {
     // This guards against the case where compositionend is triggered without the keyboard
     // (e.g. character confirmation may be done with the mouse), and keydown is triggered
     // afterwards- we wouldn't want to ignore the keydown event in this case.
-    if (this.compositionEndedAt !== null) {
-      if (Math.abs(event.timeStamp - this.compositionEndedAt) < 200) {
-        this.compositionEndedAt = null
-        return true
-      }
+    if (browser.safari && this.compositionEndedAt !== null && Math.abs(event.timeStamp - this.compositionEndedAt) < 500) {
+      this.compositionEndedAt = null
+      return true
     }
     return false
   }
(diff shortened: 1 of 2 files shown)


src/domchange.js
```diff
@@ -82,18 +83,40 @@ export class DOMChange {
...
+    // selection. On Safari, when Enter is pressed, compositionend and keydown events are
+    // emitted. The keydown event triggers newline insertion, which we don't want.
+    // This method returns true if the keydown event should be ignored.
+    // We only ignore it once, as pressing Enter a second time *should* insert a newline.
+    // Furthermore, the keydown event timestamp must be close to the compositionEndedAt timestamp.
+    // This guards against the case where compositionend is triggered without the keyboard
+    // (e.g. character confirmation may be done with the mouse), and keydown is triggered
+    // afterwards- we wouldn't want to ignore the keydown event in this case.
+    if (this.compositionEndedAt !== null) {
+      if (Math.abs(event.timeStamp - this.compositionEndedAt) < 200) {
```
marijnh: This could be a single `if` with `&&` instead of a nested one. And does it make sense to only do this check on Safari, as further protection against false positives?

kevinzang: Thanks! I've combined the nested `if`s and added Safari check. I've also increased the delta time from 200 to 500, as upon re-testing my changes I got a delta of 400 one time.

marijnh:
Thanks! Merged as 0c54477

kevinzang:
@marijnh thanks! When do new versions get created? I'm noticing the latest version on https://github.com/ProseMirror/prosemirror-view/ is 1.7.0 from 3 days ago, and I'm wondering when 1.7.1, containing this change, gets released?

marijnh:
Thanks for the reminder. I've published 1.7.1

PR #45, opened 2019-02-20
Use union for scrollThreshold and scrollMargin params

hubgit:
The `|` is causing the parameter documentation to not be parsed correctly; use `union` instead.

image


marijnh:
Thanks!

PR #46, opened 2019-02-27
Don't set contentEditable on nodeview if it's already been defined

ahixon:
Workaround for ProseMirror/prosemirror#884


marijnh:
Great, merged as 1ced0d5

PR #48, opened 2019-03-18
Always sync node selections on update state

jljorgenson18:
I'm not sure if this is the best approach but I figured I would make an attempt. We are seeing issues on Prosemirror-view 1.8.2 and 1.8.3 when deselecting node views. Recent commits stopped calling selectionToDOM when clicking off of a node view because `this.mouseDown && this.selectionReader.domChanged() && anchorInRightPlace(this)` is returning true. Although the selection is in the right place, the node selection sync does not happen which makes our node views keep their selected styles.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -150,6 +150,7 @@ export class EditorView {
       if (forceSelUpdate ||
           !(this.mouseDown && this.selectionReader.domChanged() && anchorInRightPlace(this)))
         selectionToDOM(this, false, forceSelUpdate)
+      else syncNodeSelection(this, state.selection)
       this.domObserver.start()
     }
 


cmlenz:
For what it's worth, we noticed the same problem yesterday when updating prosemirror-view from 1.7.3 to 1.8.3.

jljorgenson18:
I was going to create an issue but our setup is so custom that I didn't think we could get a repro case. This does seem like an actual bug though.

marijnh:
Thanks for looking into this---that was indeed fallout from the newly introduced kludges in 1.8.2 and 1.8.3. Your solution looks good—merging and releasing as 1.8.4

PR #49, opened 2019-03-29
Moving lastClick to the editor view to resolve reinitialization issues

jljorgenson18:
We were having a series of test failures due to the lastClick object not being reset. The tests were passing in isolation but failing when run in sequence. Turns out, the click type was being registered as a double click because of the previous lastClick object. We simulate mousedowns with the correct clientX and clientY positions which made "isNear" true. For test suites that have similar setup steps and that simulate mousedown, it would be helpful to flush lastClick when the view is reinitialized.


marijnh:
Makes sense—merged!

PR #50, opened 2019-04-08
Update domchange.js

XuXuDev:
Fix Chinese input method collaborative content is not correct

### DIFF
--- a/src/domchange.js
+++ b/src/domchange.js
@@ -293,7 +293,7 @@ function readDOMChange(view, mapping, oldState, range, allowTypeOver) {
     return
   }
 
-  let from = mapping.map(change.start), to = mapping.map(change.endA, -1)
+  let from = mapping.map(change.start), to = mapping.map(change.endA)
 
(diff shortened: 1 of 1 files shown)


XuXuDev:
Hello marijnh 
    I am a developer from China. I found a problem with a prosemirror cooperating under the Chinese input method.
    This problem will happen in this situation：
     1. Two people open this link : http://prosemirror.net/examples/collab/#edit-Example
    2.First person edit like this, but not press Space
image
    3.Another person do same thing
image
    4.First person press Space then second person press Space，we get the wrong result
image

    This is beacase domchange.js caculate "to" is wrong

marijnh:
Thanks for looking into this. Does setting `to` to `Math.max(from, mapping.map(change.endA, -1))` also solve the problem for you? I think that'd be more correct.

XuXuDev:
yes, this is also work!
thanks!

XuXuDev:
I re-submitted a Pullrequest！
Thanks for the support

PR #51, opened 2019-04-08
Fix Chinese input problem

XuXuDev:
Fix Chinese input method collaborative content is not correct

### DIFF
--- a/src/domchange.js
+++ b/src/domchange.js
@@ -293,7 +293,7 @@ function readDOMChange(view, mapping, oldState, range, allowTypeOver) {
     return
   }
 
-  let from = mapping.map(change.start), to = mapping.map(change.endA, -1)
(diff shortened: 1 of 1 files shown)


XuXuDev:
I re-submitted a PullRequest at here.
thks

XuXuDev:
Because we are all collaborative editors in Chinese.
Can U send a new version as 1.8.5?
thanks!

marijnh:
Thanks. Merged and released as 1.8.5.

As an aside, this code will be superseded again in the near future by the 1.9 release (in the `composition` branch), which fixed this issue in a different way.

XuXuDev:
Yes, we will conduct more tests in the Chinese environment, and will continue to pay attention to your progress, hope to maintain communication and make a contribution to prosemirror!

PR #53, opened 2019-06-12
parseRange: fix empty viewDesc from/to offsets

kapouer:
This is a wild guess - it fixes some issues i have had with DOMObserver, like when i insert DOM nodes in an empty contentDOM of a nodeView.
The same issue (not re-adding base to from/to) might still be present in parseRange, in other cases.


kapouer:
oops, didn't see your answer. Closing.

PR #54, opened 2019-07-19
WIP: Allow NodeViewDesc to ignore selection mutations

hubgit:
This is an implementation of [RFC 13](https://github.com/ProseMirror/rfcs/pull/13), with some caveats:

* I haven't tried it with anything other than `ignoreMutation = () => true` in a NodeViewDesc, because I didn't need to filter based on the node.
* I'm not yet sure how to write a test for this.

### DIFF
--- a/src/domobserver.js
+++ b/src/domobserver.js
@@ -84,10 +84,12 @@ export class DOMObserver {
 
   ignoreSelectionMutation(sel) {
     let desc = this.view.docView.nearestDesc(sel.anchorNode, true)
+    
+    if (!desc || !desc.node || !desc.dom) return false
 
     if (desc.node.isText) return false
 
-    return desc && desc.dom && desc.ignoreMutation({
+    return desc.ignoreMutation({
       type: 'selection',
       target: desc.dom
     })
--- a/src/domobserver.js
+++ b/src/domobserver.js
@@ -82,17 +82,11 @@ export class DOMObserver {
     this.currentSelection.set(this.view.root.getSelection())
   }
 
-  ignoreSelectionMutation(sel) {
-    let desc = this.view.docView.nearestDesc(sel.anchorNode, true)
-    
-    if (!desc || !desc.node || !desc.dom) return false
+  ignoreSelectionChange(sel) {
+    let container = sel.getRangeAt(0).commonAncestorContainer
+    let desc = this.view.docView.nearestDesc(container, true)
 
-    if (desc.node.isText) return false
-
-    return desc.ignoreMutation({
-      type: 'selection',
-      target: desc.dom
-    })
+    return desc && desc.node && !desc.node.isText && desc.node.isAtom
   }
 
   flush(mutations) {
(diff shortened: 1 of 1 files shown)


src/domobserver.js
```diff
@@ -82,6 +82,17 @@ export class DOMObserver {
     this.currentSelection.set(this.view.root.getSelection())
   }
 
+  ignoreSelectionMutation(sel) {
+    let desc = this.view.docView.nearestDesc(sel.anchorNode, true)
+
+    if (desc.node.isText) return false
```
marijnh: Why do you explicitly check for this? In any case, you probably shouldn't be dereferencing `desc` before checking whether it's null.

src/domobserver.js
```diff
@@ -82,6 +82,17 @@ export class DOMObserver {
...
+
+    return desc && desc.dom && desc.ignoreMutation({
```
marijnh: What is the `desc.dom` check for?

hubgit: The `target` property below needs to be a DOM node, so I added a check to make sure it's set - if it's guaranteed to always be there then this check can be removed.

src/domobserver.js
hubgit: Without this line selection changes in text nodes will always be ignored, presumably because of this in TextViewDesc:

```
ignoreMutation(mutation) {
    return mutation.type != "characterData"
}
```

Good point about the null check that needs to come first.

marijnh: Ah, right, we should probably change that in `TextViewDesc` rather than working around it here.

src/domobserver.js
marijnh: I guess what we really want as target (and as argument to `nearestDesc`) is the first element node that contains both the `anchorNode` and `focusNode`. (So that if one side of the selection stays in a widget, but the other escapes it, we don't ignore the selection change.)

hubgit:
I've pushed an update that finds the common ancestor node of the selection, and uses `!desc.node.isText && desc.node.isAtom` instead of calling `ignoreMutation`, as a more general rule.

Does it make sense to always ignore selection changes in atoms (except for text nodes)?

hubgit:
When the actual selection is inside a node view that's ignoring selection changes, should ProseMirror's selection be the node itself (i.e. a NodeSelection), rather than wherever it was before the ignored selection was created?

marijnh:
I was thinking more along the lines of [this branch](https://github.com/ProseMirror/prosemirror-view/tree/ignore-selection). Does that code work for you?

hubgit:
Yes, that seems reasonable. 

After testing these approaches, there are still two remaining problems:

* ProseMirror doesn't know that the selection is now inside the node that's ignoring selection changes - it still thinks the selection is wherever it was before, which is confusing when the application's UI changes based on the selection.
* If the initial click into the editor after loading the page is inside the node that's ignoring the selection change, the cursor is shown at the correct location but typing causes text to be entered at the start of the document.

marijnh:
> ProseMirror doesn't know that the selection is now inside the node that's ignoring selection changes

Right, if you want this, the node view will have to read its internal selection, translate it to an outside-editor selection, and update the outer editor's state. The CodeMirror example does this, I believe.

> If the initial click into the editor after loading the page is inside the node that's ignoring the selection change, the cursor is shown at the correct location but typing causes text to be entered at the start of the document.

Where is the focus when this happens? I'd expect it to be inside the inner editor, so that the content isn't even inserted into the outside one.

hubgit:
> Where is the focus when this happens? I'd expect it to be inside the inner editor, so that the content isn't even inserted into the outside one.

If I click inside the inner CodeMirror editor as the first action after loading the page, then `document.activeElement` is the CodeMirror `textarea`, and the cursor is flashing there, but typing causes the input to appear elsewhere, at the start of the document.

I'll try to make a minimal reproduction.

PR #55, opened 2019-07-26
Fix eq() comparison of DecorationSets

sciyoshi:
This function is currently always returning false for different objects, so the value equality comparison is not working. Seems like a typo to me, and I think this causes some unnecessary re-renders when applying the same decoration set.


marijnh:
Thanks for spotting that! It seemed the view desc reuse did prevent most DOM redrawing from taking place, but still this should definitely be an efficiency gain.

PR #57, opened 2019-08-13
Feature: Add EditorProp which allows editors to exactly determine wh…

bZichett:
…ich container is going to be scrollable

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -591,6 +591,11 @@ function changedNodeViews(a, b) {
 //   scrollMargin:: ?union<number, {top: number, right: number, bottom: number, left: number}>
 //   Determines the extra space (in pixels) that is left above or
 //   below the cursor when it is scrolled into view. Defaults to 5.
+//
+//   scrollContainerClass:: ?string
+//   Class name for the ancestor element of root editor element in control of scrollable functionality.
+//   If not set, parent nodes are recursively iterated through and a scrollable container
+//   is programmatically determined using numerous dom queries
 
 // DirectEditorProps:: interface extends EditorProps
 //
--- a/src/domcoords.js
+++ b/src/domcoords.js
@@ -11,11 +11,20 @@ function getSide(value, side) {
 }
 
 export function scrollRectIntoView(view, rect, startDOM) {
-  let scrollThreshold = view.someProp("scrollThreshold") || 0, scrollMargin = view.someProp("scrollMargin") || 5
+  let scrollThreshold = view.someProp("scrollThreshold") || 0,
+    scrollMargin = view.someProp("scrollMargin") || 5,
+    scrollContainerClass = view.someProp('scrollContainerClass')
+
   let doc = view.dom.ownerDocument, win = doc.defaultView
   for (let parent = startDOM || view.dom;; parent = parentNode(parent)) {
     if (!parent) break
     if (parent.nodeType != 1) continue
+
+    if (scrollContainerClass) {
+      const isTarget = parent.classList.contains(scrollContainerClass)
+      if (!isTarget) continue
+    }
+
     let atTop = parent == doc.body || parent.nodeType != 1
     let bounding = atTop ? windowRect(win) : parent.getBoundingClientRect()
     let moveX = 0, moveY = 0


bZichett:
@marijnh Either i'm doing something special with managing scrollable containers or there is something not exactly right with the recursive parent search for the scrollable container.  When pasting in my app, the wrong element is shifted causing white space to incorrectly appear.  I ended up testing this out in the node_modules dist and specified the className I wanted the `scrollRectIntoView` algorithm to run on and it is working perfectly.  

Has anyone else had this issue?  

Regardless of the issue, it seems the recursive search upwards operates on _all_ ancestor nodes from the source on every dispatched `scrollToSelection` so I suppose this will be more performant as it requires 1 DOM size query operation - the `getBoundClientRect` call)

src/index.js
```diff
@@ -591,6 +591,11 @@ function changedNodeViews(a, b) {
 //   scrollMargin:: ?union<number, {top: number, right: number, bottom: number, left: number}>
 //   Determines the extra space (in pixels) that is left above or
 //   below the cursor when it is scrolled into view. Defaults to 5.
+//
+//   scrollContainerClass:: ?string
+//   Class name for the ancestor element of root editor element in control of scrollable functionality.
+//   If not set, parent nodes are recursively iterated through and a scrollable container
+//   is programmatically determined using numerous dom queries
```
bZichett: Let me know if this is unclear, as it probably could be drafted better

marijnh:
Scrolling just one container isn't enough. If you have a situation like this

```
| editor
| < cursor
|
| | scrollable wrapper around editor
| |
| |
|
    | window viewport
    |
    |
    |
```

Then you'll need to scroll both the container and the body to get the cursor into view.

Can you distill the situation where scrolling goes wrong down to its minimal ingredients? Might be a bug in the scrolling code.

bZichett:
Well, I might have got myself in a special situation but here it goes

I wanted the menu bar to be positioned above the scrollable text editor so there is a wrapper element that serves as its anchor using the transform: translateZ(0) trick to do so.  (I remember at one point, ProseMirror supported this but Isomething changed and I had to work out this alternate solution. (I see there is still at least one open issue relating to menu float positioning: https://github.com/ProseMirror/prosemirror-menu/issues/18)


```md
.Document-Wrapper (Where I want the menubar to be anchored)
     .Scrollable-Container (the scrollable container)
          .ProseMirror-menubar-wrapper
             .Prosemirror-menu-bar
             .ProseMirror
```

I found that `.Document-Wrapper` requires overflow: hidden otherwise the editor scroll bar doesnt show, so this is why two elements are found by `scrollRectIntoView` and the scrollTop applies to the overflow: hidden class ( I just stepped through the code to confirm) which causes a UX / UI regression in our app.

The way I see it, given this function runs every keypress that causes text to be entered (or more?) this PR has value because it 

1) completely fixes a weird use case that I got myself into
2) reduces the amount of getBoundingClientRect() calls to 1, rather than a call for every ancestor element (my app has ~15.)  I don't know *too* much about the internals but I recall reading this function can cause browser reflows.
3) The use case you suggested seems impossible in my app.  We can 
   a) add documentation that the use case you described is not supported or
   b) add support for it by adding an array of classes? (Or however you think is best)

marijnh:
> Performance: reduces the amount of getBoundingClientRect() calls to 1, rather than a call for every ancestor element (my app has ~15.) I don't know too much about the internals but I recall reading this function can cause browser reflows.

It does require layout to be computed, but if you call it multiple times in a row it won't cause more than one reflow, so I don't think there really is much of a performance gain here.

I'm unlikely to merge this, so I'd still be interested in the setup (as code, not description) that can reproduce this.

bZichett:
> It does require layout to be computed, but if you call it multiple times in a row it won't cause more than one reflow, so I don't think there really is much of a performance gain here.

What if there is no scrollable container?  Does this mean an unnecessary browser reflow on every keypress?

marijnh:
>What if there is no scrollable container? Does this mean 1 unnecessary browser reflow on every scrollIntoView transaction?

There is going to be a browser reflow on every keypress anyway (since the document content changed, and it needs to be drawn). This happens at the end of ProseMirror's update, so it shouldn't cause any reflows that weren't going to happen anyway.

bZichett:
Thanks for the quick replies @marijnh.  Thanks for caring / not merging without thought.  Closing this as I figured out how to remove the `overflow: hidden` styles that were causing this issue to begin with.  (The algorithm found a differential on these elements when it shouldn't have due to the menubar element being crudely popped out of the editor)

bZichett:
But btw, to be clearer, although I removed the overflow: hidden styles to fix this, the bug in the scrollToRect algorithm is that it should do a `computedStyle` check on the element before applying moveX / moveY.   I can open up a PR if you are interested.

marijnh:
Good point. A PR would be great.

PR #58, opened 2019-08-24
Adding a trackingOnly check on selections before syncing prosemirror selections to the dom

jljorgenson18:
If you are ok with this, I can also submit a PR to Prosemirror-state to add this as an actual property on Selections.

What we are trying to do is have a custom selection that is tracked in Prosemirror state but is "hands off" when it comes to the native selection. We have a few NodeViews where we want to give users the ability to select text and interact with the NodeViews in a few different ways. The problem is, what should the Prosemirror selection be when the native selection is inside one of these NodeViews?

We built a new custom selection called NestedSelection and added a selection listener in a plugin. If the selection is inside one of these special containers, we set the NestedSelection in a similar way to how we set NodeSelection. This works just fine after the selection is set, but during the initial setting of NestedSelection, prosemirror-view calls selectionToDOM and overwrites the native selection once we start selecting content.

Since the NodeView is essentially an atom, we don't get access to the `child.setSelection` condition in `ViewDesc.setSelection`. However, if we skip the `docView.setSelection` call in selectionToDOM, everything works beautifully and the interaction appears to be flawless.

So because of this, I'd like to propose a new field that allows selections to be tracked from the native selection, but does not force the prosemirror selection to update the native selection.

### DIFF
--- a/src/selection.js
+++ b/src/selection.js
@@ -47,17 +47,20 @@ export function selectionToDOM(view, takeFocus, force) {
   if (view.cursorWrapper) {
     selectCursorWrapper(view)
   } else {
-    let {anchor, head} = sel, resetEditableFrom, resetEditableTo
-    if (brokenSelectBetweenUneditable && !(sel instanceof TextSelection)) {
-      if (!sel.$from.parent.inlineContent)
-        resetEditableFrom = temporarilyEditableNear(view, sel.from)
-      if (!sel.empty && !sel.$from.parent.inlineContent)
-        resetEditableTo = temporarilyEditableNear(view, sel.to)
-    }
-    view.docView.setSelection(anchor, head, view.root, force)
-    if (brokenSelectBetweenUneditable) {
-      if (resetEditableFrom) resetEditableFrom.contentEditable = "false"
-      if (resetEditableTo) resetEditableTo.contentEditable = "false"
+    let {anchor, head} = sel
+    if(!sel.trackingOnly) {
+      let resetEditableFrom, resetEditableTo
+      if (brokenSelectBetweenUneditable && !(sel instanceof TextSelection)) {
+        if (!sel.$from.parent.inlineContent)
+          resetEditableFrom = temporarilyEditableNear(view, sel.from)
+        if (!sel.empty && !sel.$from.parent.inlineContent)
+          resetEditableTo = temporarilyEditableNear(view, sel.to)
+      }
+      view.docView.setSelection(anchor, head, view.root, force)
+      if (brokenSelectBetweenUneditable) {
+        if (resetEditableFrom) resetEditableFrom.contentEditable = "false"
+        if (resetEditableTo) resetEditableTo.contentEditable = "false"
+      }
     }
     if (sel.visible) {
       view.dom.classList.remove("ProseMirror-hideselection")


marijnh:
Doesn't providing a (no-op) [`setSelection`](http://prosemirror.net/docs/ref/#view.NodeView.setSelection) function for your nodeview handle this?

marijnh:
Where is the document focus when the user is working with these node views? Since you say 

> The problem is, what should the Prosemirror selection be when the native selection is inside one of these NodeViews?

If it is _inside_ the node view, that sounds like it has (or should have) focus, in which case ProseMirror will not mess with the selection. If the selection is a regular node selection and ProseMirror is focused, then maintaining its usual mapping between its selection state and DOM selection seems appropriate.

jljorgenson18:
I think to get around this, I'm just going to unset what Prosemirror does by restoring the range. Here is what we roughly have inside of the view function.

```typescript
 const handleSelectionChange = (event: Event): void => {
                const sel = window.getSelection();
                const nestedSelectable = maybeGetNestedSelectableBlock(sel);
                if (nestedSelectable) {
                    const pos = view.posAtDOM(nestedSelectable, 0, -1);
                    const newSelection = NestedSelection.create(view.state.doc, pos);
                    if (!newSelection.eq(view.state.selection)) {
                        try {
                            const range = sel.getRangeAt(0); // Restore the previous range since Prosemirror will try to sync the selection to dom
                            view.dispatch(view.state.tr.setSelection(newSelection));
                            sel.removeAllRanges();
                            sel.addRange(range);
                        } catch (err) {
                            log.error('Error setting nested selection');
                            log.error(err);
                        }
                    }
                }
            };
            document.addEventListener('selectionchange', handleSelectionChange);
```

This seems to work so I can close the PR if this is not a direction Prosemirror would want to go down. However, I do think implementing a "NestedSelection" inside of Prosemirror would make the selection state a bit more deterministic and could help.

marijnh:
Nested selection doesn't feel like a great abstraction, and I can imagine problems coming up when the DOM selection is somehow moved out of the node view in an unanticipated way, but the nested selection continues to leave it where it is.

A way for selections to completely override the way the DOM selection is set might be useful. But we have to go through the [RFC](https://github.com/prosemirror/rfcs/) process to add new features. I'd do this in the form of a method on the Selection class that is called instead of the block that you update in this PR, if present. Would you be willing to write up the RFC document?

jljorgenson18:
> Nested selection doesn't feel like a great abstraction, and I can imagine problems coming up when the DOM selection is somehow moved out of the node view in an unanticipated way, but the nested selection continues to leave it where it is.

Would you care to elaborate on this? Moving the DOM selection out of the Node view shouldn't cause any problems since it wouldn't be any different than a native selection being moved. The only big issues I see is future multi range implementations. 

The abstraction is mainly there so that Prosemirror itself does not get out of sync with the native selection inside of NodeViews and so we have control to prevent commands and change the style of a NodeView if you are inside of it. The NestedSelection on the Prosemirror side would be equivalent to a NodeView in terms of positioning so I'm not sure what problems pop up.

I can talk with the team and see if I should submit an RFC. I appreciate you looking at this.

jljorgenson18:
I've been giving this more thought and I think I'm just going to decline this PR and not do an RFC. We have a solution in place by undoing Prosemirror's selectionToDOM syncing and the NestedSelection works just fine now. Because this is a pretty rare case and because I don't really see other great use cases for allowing Selections to set their own native selection, I can't really write up an RFC. The only strong argument I have for this is a NestedSelection and creating a no-op.

All of that being said, NestedSelections are working great for us now and we haven't had any problems. It's been great to know when content inside of NodeViews are being selected and to make sure we disable commands and update styles accordingly. If you want any additional information on it or if you want to see the plugin, let me know and I can submit something in the discussion forum.

PR #60, opened 2019-09-24
Check observer existence in domobserver methods

ranmocy:


ranmocy:
Updated for flush().

Not in production, but in tests with jsdom framework. It's undefined function issue since `this.observer.takeRecords()` is called but `this.observer` is undefined.

jsdom added the support of mutationobserver in https://github.com/jsdom/jsdom/issues/639, and I'm waiting for next version of jest to include the [version bump](https://github.com/facebook/jest/pull/8851).

Even though it will eventually work, I still think it's better to do so for correctness. The observer should either be nullable in the caller, or silently no-op when methods are called.

src/domobserver.js
```diff
@@ -121,7 +121,7 @@ export class DOMObserver {
   }
 
   flush() {
-    if (!this.view.docView || this.flushingSoon) return
+    if (!this.view.docView || this.flushingSoon || !this.observer) return
```
marijnh: I think it'd make more sense to put the check in the next statement—it's possible for there to be events on the queue (via the chardata event handler) even when there isn't a mutation observer, and selection changes are also flushed through here.

ranmocy: Done.

marijnh:
Thank you! Merged as f329f82

PR #61, opened 2019-09-25
Add documentation for undocumented parameter on node().

TkTech:
I hope this is correct, doesn't seem to have any keys with side effects like inline, just used for equality checks.


marijnh:
Looks good, thanks!

PR #62, opened 2019-10-14
Add a warning when prosemirror.css has not been loaded.

hitsthings:
Adds the warning requested in #981.

I'll let you decide whether to merge it. I personally think it's helpful functionality, but
1) it's not easily testable in your unit test framework (because I'd have to mock both console.warn and  the "unloaded" state of prosemirror.css)
2) It requires a custom exception in blint to allow us to use the console. (https://github.com/ProseMirror/prosemirror/pull/987)


marijnh:
I agree this feels a bit heavy. What do you think about attached patch?

hitsthings:
Yeah way fewer gymnastics. I think that'll be really helpful, thanks!

marijnh:
Okay, I'll go with that then.

PR #64, opened 2019-11-15
Tweaking patchAttributes to allow for the removal of css custom properties

jljorgenson18:
Using the style setter is not sufficient when dealing with custom properties so this fix uses the CSS Declaration removeProperty instead. Otherwise, styles aren't removed on Node decorations.

Also, tests were not running due to a relative import to Prosemirror-state outside of the repo so I imported from Prosemirror-state directly.


jhnsnc:
Any updates on this?

We're starting to use CSS variables in a few places and this PR looks like it would help ensure the expected behavior around Node decorations and CSS variables.

marijnh:
Sorry, not sure how I missed this the first time around. Merged now.

PR #65, opened 2019-11-17
Firefox: eliminate bogus BR appeared inside of LI in case of backspacing last character in list item

matolg:
**Prosemirror-view version:** 1.11.2
**FF version:** 69.0.3, 69.0.6, 70.0.1
**FF issue:** https://bugzilla.mozilla.org/show_bug.cgi?id=1596374

**Steps to repro:**
1. Create a list inside the editor each item starts with SPAN or custom Node and have some text after it.
2. Put cursor at the end of any list item.
3. Start delete text until it reaches Node.
4. Observe the result.

**Expected:**
- Last text character deleted and cursor reached DOM element without visible changes in list structure.

**Actual:**
- Blank line after the list item appeared breaking list on to two parts visually.

**NB:**
- This is basically FF issue, and corresponding bug filed there. This PR is to compensate that issue considering it is not clear how long it will take to have a fix on FF side.
- I'm not very familiar with PM, so it might be better way to fix this

**Demo in PM editor:**
Kapture 2019-11-17 at 14 48 21

### DIFF
--- a/src/domobserver.js
+++ b/src/domobserver.js
@@ -153,7 +153,7 @@ export class DOMObserver {
     // Firefox adds bogus BR to the very end of LI node
     // in case of backspacing last character in the LI and having inline DOM element in front of it (e.g. SPAN or custom TODO)
     // FF issue: https://bugzilla.mozilla.org/show_bug.cgi?id=1596374
-    if (result.gecko &&
+    if (browser.gecko &&
         mutations.some(mut =>
           mut.target &&
           mut.target.nodeName === "LI" &&
--- a/src/domobserver.js
+++ b/src/domobserver.js
@@ -149,6 +149,16 @@ export class DOMObserver {
         }
       }
     }
+
+    // Firefox adds bogus BR to the very end of LI node
+    // in case of backspacing last character in the LI and having inline DOM element in front of it (e.g. SPAN or custom TODO)
+    // FF issue: https://bugzilla.mozilla.org/show_bug.cgi?id=1596374
+    if (result.gecko &&
+        mutations.some(mut =>
+          mut.target.nodeName === "LI" &&
+          mut.addedNodes.length === 1 &&
+          mut.addedNodes[0].nodeName === "BR")) to += 2
+
     if (from > -1 || newSel) {
       if (from > -1) {
         this.view.docView.markDirty(from, to)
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks for researching this so thoroughly. I'm a bit confused by the patch—I can't imagine `result.gecko` ever being true, so it looks like the new code will never fire. Are you sure this helps for you?

matolg:
Hi. Yes, I've tested this fix and it was working in ff. And `result.gecko` definitelly was true while the debuging..

marijnh:
Ah, I see, `result` refers to the object imported as `browser`. The `result` name isn't actually available in this module, so I'm still not sure how the patch could work, as it is.

What is the idea behind adding 2 to `to`? Why does that solve the issue?

matolg:
Ah, yes, now I see, what you was concerned about. `result` is a name that was in compiled version of PM, where I've done experiments, so I c/p it automatically. Fixed

marijnh:
Thanks. Could you address my second question—on why you update `to` like that?

matolg:
Yep. So if I understand it correctly, extending range leads to marking whole `<li>` as dirty instead of paragraph inside. `<br>` is added on the top level of `<li>` making it dirty. So marking it as dirty leads to recreation it as whole but without that `<br>`

this also could be fixed like this:
```javascript
...
  var extendRange = browser.gecko &&
      mutations.some(mut =>
        mut.target &&
        mut.target.nodeName === "LI" &&
        mut.addedNodes.length === 1 &&
        mut.addedNodes[0].nodeName === "BR")

  if (from > -1 || newSel) {
    if (from > -1) {
      this.view.docView.markDirty(from, extendRange ? to + 2 : to);
...
```

matolg:
Hi. Any updates on this?

matolg:
Can't we accept this fix?

marijnh:
No, it is really too arbitrary a kludge. There are schemas in which BR nodes do appear directly in LI nodes, so treating that as an indication of this bug isn't very robust, and relying on the DOM parser to remove that second BR node isn't guaranteed to work. Adding two to `to` might even put it out of range of the current document, which will probably crash the editor.

I've implemented an alternative workaround in attached patch. Can you confirm whether it works for you?

matolg:
Hi. I've tested this with last version (1.13.6) and it works fine. Thanks for response. I think, we can close this pr then

PR #66, opened 2019-12-19
Not registering style mutations if the value did not change at all

jljorgenson18:
We were seeing infinite loops from our forecolor mark on Chrome because Chrome was registering that the style value had changed, even though it did not. This then triggered our forecolor parse rule which then caused the mark to serialize to dom which then triggered another mutation. This PR skips style mutations if the value doesn't actually at all.

### DIFF
--- a/src/domobserver.js
+++ b/src/domobserver.js
@@ -176,8 +176,12 @@ export class DOMObserver {
     let desc = this.view.docView.nearestDesc(mut.target)
     if (mut.type == "attributes" &&
         (desc == this.view.docView || mut.attributeName == "contenteditable" ||
-         // Firefox sometimes fires spurious events for null/empty styles
-         (mut.attributeName == "style" && !mut.oldValue && !mut.target.getAttribute("style"))))
+         (mut.attributeName == "style" && 
+          // Firefox sometimes fires spurious events for null/empty styles
+          (!mut.oldValue && !mut.target.getAttribute("style")) ||
+          // Chrome sometimes fires mutations for styles, even though the value doesn't change at all.
+          // This can cause infinite loops as styles are reparsed
+          mut.oldValue == mut.target.getAttribute("style"))))
       return null
     if (!desc || desc.ignoreMutation(mut)) return null
 


jljorgenson18:
Just to add a bit more detail, the infinite loops were happening during compositions. There might be a better way to handle but this did seem specific to styles when I was debugging. Other attributes did not seem to have the same behavior.

marijnh:
What kind of transactions were being generated in the infinite loop? It seems like if there's no actual change, the DOM change handler should end up comparing two identical documents and not firing a transaction. Where did that go wrong?

jljorgenson18:
I can dig into this a bit more today but my guess is that the diff  isn't working correctly due to some special cased composition handling. The transactions that were firing were constantly replacing the current composition text over and over in the document. This also seemed to be specific to Japanese, but I'd have to double check other IMEs.

In general though, is this not the right spot for this fix? It sounds like the underlying problem is in the document diffing.

jljorgenson18:
We had a similar infinite loop on Android due to us dispatching meta transactions for tracking compositions which caused an issue with the cursorwrapper. We had to add a getter and setter to "freeze" the cursorwrapper so it wasn't unset when we dispatched (although eventually we just disabled the markCursor altogether on Android). I'll dig into that as well to see if that is playing a role.

jljorgenson18:
That Android cursorWrapper issue does not seem relevant at all so we could ignore that.

It looks like there are three key pieces.

1. The actual registered mutation
2. Finding the diff when parsing from dom
3. Updating the dom from updating the state (which then triggers the mutation)

The registered mutation could be expected. Even though the value doesn't change, the style attribute may still be with the same value.

As for the diff when parsing, it looks like an actual diff is returned here

```js
let change = findDiff(compare.content, parse.doc.content, parse.from, preferredPos, preferredSide)
```

After typing with Japanese with a forecolor stored mark, I see this result from findDiff

`{start: 1, endA: 2, endB: 2}`

Despite seeing the same thing on "a" and "b". This is what I get when logging toDebugString on both

`<forecolor("み")> <forecolor("み")>`

Once the diff is found, the text is replaced with the same exact text here

```js
else if ($from.parent.child($from.index()).isText && $from.index() == $to.index() - ($to.textOffset ? 0 : 1)) {
      // Both positions in the same text node -- simply insert text
      var text = $from.parent.textBetween($from.parentOffset, $to.parentOffset);
      if (view.someProp("handleTextInput", function (f) { return f(view, chFrom, chTo, text); })) { return }
      tr = view.state.tr.insertText(text, chFrom, chTo);
    }
```

I'm still trying to figure out how the whole "dirty" mechanism works but somehow when updating the state, the mark is mutated.

That leads to three possible solutions

1. Don't register attribute changes that don't actually change anything
2. Tweaking the initial parse diffing
3. Tweak the reconciliation between the state to dom.

I'm not sure where the fix should be, but not registering mutations could be the easiest win because it shortcuts having to do any diffs or dirty checks. Then again, I don't know if anything unforeseen would happen.

Also, our forecolor mark toDOM method looks like this

```js
toDOM: mark => {
            const {
                attrs: { color, inversionType }
            } = mark;
            const darkModeColor = invertColor(color, {
                attribute: 'color',
                inversionType: inversionType
            });
            return [
                'span',
                {
                    style: `--darkmode-color: ${darkModeColor}; --lightmode-color: ${color};`,
                    class: forecolorClass
                }
            ];
        }
```

And there is a parse rule to reparse this output as needed

```js
    {
        tag: '.' + forecolorClass,
        context: 'inlinecontainer/',
        getAttrs(node: HTMLSpanElement) {
            const color = node.style.getPropertyValue('--lightmode-color');
            return color && color !== 'inherit' ? { color } : false;
        }
    }
```

marijnh:
`findDiff` in domchange.js looks like it should be returning `null` for identical inputs. And it doesn't look like the `typeOver`-related logic after the call to `findDiff` is kicking in here (though if you could confirm that that'd be useful). What does `parse.doc.content.eq(compare.content)` return? The `toString` output doesn't include attributes, so maybe there's some difference hidden in there?

jljorgenson18:
So it looks like there was a difference in attributes on the forecolor mark which was causing a diff to return. I did fix this to make sure that findDiff was returning null. However, the endless loop is still there and it seems to be due to the browser just being confused. This happened on Chrome, Firefox, and Safari and it was realllllly bad on Safari.

Looking at the trace, it seems like it could be an OS issue. I don't see where the mutation is occurring and the stack doesn't show past the initial mutation registration.

I logged the trace here in findDiff

```
  var change = findDiff(compare.content, parse.doc.content, parse.from, preferredPos, preferredSide);
  console.trace('Change???', change, parse.doc.content.eq(compare.content));
```

Which then successfully returned null after my change. I then see this in the logs.

Screen Shot 2020-01-02 at 4 45 02 PM

Which shows findDiff firing and over. The stacktrace just shows flush so it looks like it is coming directly from the browser.

marijnh:
Weird. Are any transactions firing during the endless loop?

jljorgenson18:
Nope. We have a transaction logger and you can see in the image of the logs that "TR sel: 2" represents a transaction that set a text selection at position 2. There were no other transactions between that selection transaction and the endless loop. `findDiff` returns nulls and nothing occurs.

I think in order to proceed, I can try to repro this with the prosemirror demo app and do a bit more research unless you have anything else I can try.

marijnh:
No, I agree, a minimal reproduction seems a good way forward.

jljorgenson18:
I found the culprit and it was something specific to our setup... but maybe we can do something inside of Prosemirror to help.

We have utility method that parses attributes from the dom and will look at the style string for certain attributes. This utility was calling `dom.setAttribute('style', someStyleString)`, even though the value of the style was not changing. This caused the infinite loop as our parse rules were causing a mutation to occur, even though the actual value was not changing. I refactored the code a bit and the infinite loop is now gone.

This was incredibly difficult to pinpoint and I'm wondering if it would be beneficial to ignore mutation that occur while Prosemirror is parsing the dom. At the very least, we could potentially log a warning or somehow prevent infinite loops. I'd be happy to help add safety checks for accidental dom mutations.

marijnh:
It does seem a bit unusual for _parsing_ logic to mutate the DOM. I guess it would be possible to detect too many DOM change events occurring right after each other, but the code for that wouldn't be all that pretty, and the only thing it'd be able to do would be to break the cycle and print a warning, which doesn't help that much with debugging the source of the problem.

It's unfortunate that there's no good way to go from a DOM mutation to the code that caused it, but dev tools on modern browsers allow you to set breakpoints on DOM mutation, so it's not a huge problem to go from the diagnosis 'something keeps mutating the DOM' to finding the code responsible.

On the whole, I'm leaning towards not adding special code for this.

jljorgenson18:
I should have tried "break on subtree modification" sooner.

Detection code would be difficult, but I was also thinking of just disconnecting the mutation observer during parse operations. However, this does seem rare enough where the cost of bad looking code is probably too high.

I would say that it is definitely unusual but not unfounded. We weren't intentionally mutating the dom and the actual value was not changing. (Side note, we had to mutate the style value directly because it was essentially there to correct old invalid css properties. It was the easiest way to make it work). 

I'll go ahead and close this and can follow up if I discover anything new.

PR #69, opened 2020-02-18
fix wierd compositioning of multibyte language.

rubang:
This patch fixes an issue reported in https://github.com/ProseMirror/prosemirror/issues/1032

### DIFF
--- a/src/domobserver.js
+++ b/src/domobserver.js
@@ -106,7 +106,7 @@ export class DOMObserver {
     if (browser.ie && browser.ie_version <= 11 && !this.view.state.selection.empty) {
       let sel = this.view.root.getSelection()
       // Selection.isCollapsed isn't reliable on IE
-      if (sel.focusNode && isEquivalentPosition(sel.focusNode, sel.focusOffset, sel.anchorNode, sel.anchorOffset))
+      if (!this.view.composing && sel.focusNode && isEquivalentPosition(sel.focusNode, sel.focusOffset, sel.anchorNode, sel.anchorOffset))
         return this.flushSoon()
     }
     this.flush()
--- a/src/domobserver.js
+++ b/src/domobserver.js
@@ -34,7 +34,6 @@ export class DOMObserver {
     this.view = view
     this.handleDOMChange = handleDOMChange
     this.queue = []
-    this.flushingSoon = false
     this.observer = window.MutationObserver &&
       new window.MutationObserver(mutations => {
         for (let i = 0; i < mutations.length; i++) this.queue.push(mutations[i])
@@ -61,10 +60,7 @@ export class DOMObserver {
   }
 
   flushSoon() {
-    if (!this.flushingSoon) {
-      this.flushingSoon = true
-      window.setTimeout(() => { this.flushingSoon = false; this.flush() }, 20)
-    }
+    window.setTimeout(() => { this.flush() }, 20)
   }
 
   start() {
@@ -80,7 +76,7 @@ export class DOMObserver {
       let take = this.observer.takeRecords()
       if (take.length) {
         for (let i = 0; i < take.length; i++) this.queue.push(take[i])
-        window.setTimeout(() => this.flush(), 20)
+        this.flushSoon()
       }
       this.observer.disconnect()
     }
@@ -131,7 +127,7 @@ export class DOMObserver {
   }
 
   flush() {
-    if (!this.view.docView || this.flushingSoon) return
+    if (!this.view.docView) return
     let mutations = this.observer ? this.observer.takeRecords() : []
     if (this.queue.length) {
       mutations = this.queue.concat(mutations)
--- a/src/viewdesc.js
+++ b/src/viewdesc.js
@@ -218,7 +218,7 @@ class ViewDesc {
       let desc = this.getDesc(cur)
       if (desc && (!onlyNodes || desc.node)) {
         // If dom is outside of this desc's nodeDOM, don't count it.
-        if (first && desc.nodeDOM && !(desc.nodeDOM.nodeType == 1 ? desc.nodeDOM.contains(dom) : desc.nodeDOM == dom)) first = false
+        if (first && desc.nodeDOM && dom.nodeValue != '' && !(desc.nodeDOM.nodeType == 1 ? desc.nodeDOM.contains(dom) : desc.nodeDOM == dom)) first = false
         else return desc
       }
     }


marijnh:
That code to force a delay of the dom flush isn't there for the fun of it. It fixes issues like prosemirror/prosemirror#967, and as such should not be removed.

rubang:
Thanks for your reply.
I didn't removed a delay.
I just removed the `flushingSoon` flag.
Without a proper patch, typing in Korean is really annoying. 
This issue was also reported in discuss.prosemirror.net but no updates during a couple of months.
https://discuss.prosemirror.net/t/problems-with-replacing-non-empty-selection-in-ie11/2180/6
BTW, I will check whether the issue you attached might be reproduced by my patch.

marijnh:
> I just removed the flushingSoon flag.

Right, but that is necessary to 'batch' flushes together, by holding them up until the delay has elapsed.

marijnh:
(If you can provide your diagnosis of the issue here, maybe we can find a solution that doesn't have this problem.)

rubang:
ProseMirror/prosemirror#967 issue is not reproduced by my patch.

Without removing `flushingSoon' flag as you suggested, I think we have to ignore a mutation of EmptyTextNode in IE.
When I change MutationObeserver to ignore EmptyTextNode mutation like below,
composition of first character completes well.
`this.observer = window.MutationObserver &&
      new window.MutationObserver(mutations => {
        for (let i = 0; i < mutations.length && mutations[i].target.nodeValue != ''; i++) // changed
          this.queue.push(mutations[i])
        
        // IE11 will sometimes (on backspacing out a single character
        // text node after a BR node) call the observer callback
        // before actually updating the DOM, which will cause
        // ProseMirror to miss the change (see #930)
        if (browser.ie && browser.ie_version <= 11 && mutations.some(
          m => m.type == "childList" && m.removedNodes.length ||
               m.type == "characterData" && m.target.nodeValue != '' && m.oldValue.length > m.target.nodeValue.length)) // changed
          this.flushSoon()
        else
          this.flush()
      })
    this.currentSelection = new SelectionState
    if (useCharData) {
      this.onCharData = e => {
        if (e.target.nodeValue != '') { // changed
          this.queue.push({target: e.target, type: "characterData", oldValue: e.prevValue})
          this.flushSoon()
        }
      }
    }`

When typing two characters sequentially during composition, mutation.target always includes EmptyTextNode. But I don't know for sure this will do the trick for every cases.

marijnh:
I've pushed a few patches (212454bf28e3892...dc1c0de0f93161e71) that seem to make the problem less bad when I test it. But I don't know the first thing about the Korean script or the way Korean IME is supposed to work, so please take a look and see whether it helps at all for you.

MoonSupport:
Thank you for your interest in this issue.

Your work solves one of my problems!

I can write the first letter in Korean!!!

But the last letter is still removed.

And unfortunately there's another issue... 

It is problem that enter duplicate characters when i input lower consonant composition called **Batchim**

<h3>It works</h3>

---

works

Row1 - The **Batchim** (‘ㄴ’) of the first letter and the consonant (‘ㄴ’) of the second letter cannot be combined.

Row2 and Row3 works by the same reason as Row1.

Row4 - Although the **Batchim** ('ㄱ') of the first letter  and the consonant (‘ㄴ’) of the second letter can be combined, it works if the two consonants are the same.

<h3>work problem</h3>

---

dontwork

Row1. If there is no **Batchim** in first letter, the second consonant (‘ㅇ’) is duplicated as the **Batchim**. 
( Expected : 아이스, Result : 앙이스 ) - insert twice 'ㅇ'

Row2. When there is **Batchim** (‘ㄴ’) in the first letter, the second letter consonant('ㅈ') is duplicated into the consonant below the first letter if it can be combined with the second letter.
( Expected : 연주, Result : 엱주 ) - insert twice 'ㅈ'

If you need more information, Please comment to me anytime.

marijnh:
You're going to have to describe the problem in a way that tells me precisely the (minimal) keys I have to press with Win 10's Korean IME on a Latin keyboard, and the characters that should be produced by those—I don't have the background knowledge to parse your current description of the problem.

MoonSupport:
I hope the description below is what you want

Here are two cases
<h3>1. 아이</h3>

Press
Latin Keyboard - d + k + d + l 
Korean - ㅇ + ㅏ + ㅇ + ㅣ

Expected
Latin Keyboard - dkdl 
Korean - 아이

Result
Latin Keyboard - dkddl ( twice 'd' ) 
Korean - 앙이 ( twice 'ㅇ ) 

<h3>2. 연주</h3>

Press
Latin Keyboard - d + u + s + w + n 
Korean - ㅇ + ㅕ + ㄴ + ㅈ + ㅜ

Expected
Latin Keyboard - duswn 
Korean - 연주

Result
Latin Keyboard - duswwn ( twice 'w' ) 
Korean - 엱주 ( twice 'ㅇ ) 

This only occurs on the first letter.

I hope this explanation is helpful to you

rubang:
@marijnh I tested your patches (212454b...dc1c0de).
There are two issues in these patches.

1. First issue (well described by @MoonSupport) is about composition of the very first character in Korean. I tested a few things based on your patches. I found some solution but I can't be sure if there would be other relevant issues.

in src/viewdesc.js, I commented out the second line.
```js
export function endComposition(view, forceUpdate) {
  view.composing = false
  // while (view.compositionNodes.length > 0) view.compositionNodes.pop().markParentsDirty()
  if (forceUpdate || view.docView.dirty) {
    let sel = selectionFromDOM(view)
    if (!sel.eq(view.state.selection)) view.dispatch(view.state.tr.setSelection(sel))
    else view.updateState(view.state)
    return true
  }
  return false
}
```
I think the `markParentsDirty` function have some problems.
```
markParentsDirty() {
    let level = 1
    for (let node = this.parent; node; node = node.parent) {
      let dirty = level == 1 ? CONTENT_DIRTY : CHILD_DIRTY // <- It is same with `let dirty = CONTENT_DIRTY`
      if (node.dirty < dirty) node.dirty = dirty
    }
  }
```
I don't know the purpose of this function but when I removed the code to call this function, compositions of every characters worked fine.

2. Second issue is about when pressing enter during composition of a character, editor removes the character.
Feb-25-2020 13-00-13

This issue is produced by this patch https://github.com/ProseMirror/prosemirror-view/commit/e939ef72164fcb4125d1566e56394be0ee748fcb (made to solve ProseMirror/prosemirror#967 issue)
Selection change made by mouse double clicking and selection change made by typing and composition are not same. That's why I made this commit to solve this issue.
https://github.com/ProseMirror/prosemirror-view/pull/69/commits/def6de467df29618711358e6a4277cf1e179830e

pawsong:
@marijnh I really appreciate you that you made the best web editor in the open source society. 

Would you give some suggestion so that we can help this issue go forward?

If [the information](https://github.com/ProseMirror/prosemirror-view/pull/69#issuecomment-589229237) @MoonSupport provided is not enough, we can do additional job for it.

If the changes @rubang made [here](https://github.com/ProseMirror/prosemirror-view/pull/69#issuecomment-590673391) is acceptable, then we can make another PR for it.

Thank you in advance :)

marijnh:
I don't always have time to respond to complicated stuff like this right away. This issue is on my list.

marijnh:
I can confirm that there was a bug in `markParentsDirty` (fixed in 263d693). Does that help with the first issue?

For the second issue, patch 95d5680 might help—the problem was that enter triggered a command that was executed while the DOM changes hadn't been processed yet, and was thus looking at the wrong selection.

pawsong:
@marijnh master branch works fine. Both issues are resolved. Thank you!

marijnh:
Glad to hear! I'll do a release of prosemirror-view tomorrow.

marijnh:
Released 1.14.3

pawsong:
@marijnh Thank you!

rubang:
Everything works fine. @marijnh Thank you!

PR #70, opened 2020-03-16
Fix coordsAtPos after line wrap point

petermatik:
Coordinates on Firefox and Safari are invalid after line wrap point.

pm-coordsAtPos-linewrap

Webkit seems to support querying empty text ranges and for Firefox we need to
compare before and after top positions to get correct coordinates.

### DIFF
--- a/src/domcoords.js
+++ b/src/domcoords.js
@@ -283,16 +283,18 @@ export function coordsAtPos(view, pos) {
   let {node, offset} = view.docView.domFromPos(pos)
 
   // These browsers support querying empty text ranges
-  if (node.nodeType == 3 && (browser.chrome || browser.gecko)) {
+  if (node.nodeType == 3 && browser.webkit) {
     let rect = singleRect(textRange(node, offset, offset), 0)
     // Firefox returns bad results (the position before the space)
     // when querying a position directly after line-broken
     // whitespace. Detect this situation and and kludge around it
     if (browser.gecko && offset && /\s/.test(node.nodeValue[offset - 1]) && offset < node.nodeValue.length) {
       let rectBefore = singleRect(textRange(node, offset - 1, offset - 1), -1)
-      if (Math.abs(rectBefore.left - rect.left) < 1 && rectBefore.top == rect.top) {
+      if (rectBefore.top == rect.top) {
         let rectAfter = singleRect(textRange(node, offset, offset + 1), -1)
-        return flattenV(rectAfter, rectAfter.left < rectBefore.left)
+          if (rectAfter.top != rect.top) {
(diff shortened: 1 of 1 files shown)


src/domcoords.js
```diff
@@ -283,16 +283,18 @@ export function coordsAtPos(view, pos) {
   let {node, offset} = view.docView.domFromPos(pos)
 
   // These browsers support querying empty text ranges
-  if (node.nodeType == 3 && (browser.chrome || browser.gecko)) {
+  if (node.nodeType == 3 && browser.webkit) {
```
marijnh: You're turning off this branch for Gecko, which would make the `if (browser.gecko ...)` condition below always false, no?

src/domcoords.js
```diff
@@ -283,16 +283,18 @@ export function coordsAtPos(view, pos) {
...
     let rect = singleRect(textRange(node, offset, offset), 0)
     // Firefox returns bad results (the position before the space)
     // when querying a position directly after line-broken
     // whitespace. Detect this situation and and kludge around it
     if (browser.gecko && offset && /\s/.test(node.nodeValue[offset - 1]) && offset < node.nodeValue.length) {
       let rectBefore = singleRect(textRange(node, offset - 1, offset - 1), -1)
-      if (Math.abs(rectBefore.left - rect.left) < 1 && rectBefore.top == rect.top) {
+      if (rectBefore.top == rect.top) {
```
marijnh: The check on the `left` property is dropped here. Is that intentional? Why?

src/domcoords.js
petermatik: Right. I have access only to Firefox 70.0.1 at that version is showing browser.webkit true on Win7 VM. Nevertheless, we should indeed add gecko browser condition here.

src/domcoords.js
petermatik: I was addressing only issue with `top` and `bottom` coordinates which in this combination didn't pass this condition. `Math.abs(rectBefore.left - rect.left)` on Firefox (70.0.1) was always approx 5px.
Do you know the reason why we need to check `left` property as well?

src/domcoords.js
marijnh: Ah, lovely, Firefox has implemented the webkit-prefixed style I used for browser detection. Going to have to come up with something else there.

src/domcoords.js
marijnh: I believe it is there to make sure the rect for the position before the target is an actual unbroken, cursor-shaped rectangle, but not entirely sure. I'll play around with this ab bit more to see what the rectangles actually look like.

marijnh:
Removing the `left` check seems unproblematic. I've merged this (with a few formatting fixes) as e8b45be

PR #71, opened 2020-04-13
Use fallback in setSelection to fix chrome bug

brentropy:
In some cases with Chrome the selection is empty after calling collapse, even when it should be valid. This appears to be a bug, but it is difficult to isolate. If this happens fallback to the old path without using extend.

See [discussion](https://discuss.prosemirror.net/t/domexception-in-chrome/2613)


brentropy:
Thanks @marijnh

PR #72, opened 2020-04-14
clarify type for handlePaste event arg

maxlever:
# Purpose
clarify type of event arg to [Clipboard event](https://developer.mozilla.org/en-US/docs/Web/API/ClipboardEvent/ClipboardEvent) rather than a generic event, since this event should have a `clipboardData` property. This fixes some typescript issues.

please see https://github.com/DefinitelyTyped/DefinitelyTyped/pull/43904 for typescript fix.


marijnh:
Thanks!

PR #73, opened 2020-06-01
Moving safari nonsense check to place where elt existence ensured

alexkonstantinovakvelon:
Safari: in some cases posAtCoords throws an error `TypeError: null is not an object (evaluating 'elt.draggable')`.
This is caused by safari nonsense check in place where elt may be null and breaks position retrieval.

**Solution** - moving safari nonsense check to place where elt existence ensured.


marijnh:
Good call. Merged!

PR #74, opened 2020-06-01
FIX cross inline node decoration

ibone:
If a decoration start in a inline node ,and end in other text node. the decoration will go wrong

### The sample:
tag `<a>` schema spec type is a inline node(not a mark)
tag `<p>` schema spec type is a block node
decoration from:`13` to:`21`

`<p>normal text<a href="http://foo.bar">LINK</a>normal text</p>`

### Before fix
Screen Shot 2020-06-01 at 7 31 18 PM
### After fix
Screen Shot 2020-06-01 at 7 30 49 PM

spend several hours, i found the reason that the decoration active state was lost after `iterDeco` in child inline node

my English is poor,  Hope it helps


marijnh:
Thanks!

PR #75, opened 2020-06-08
Tweaking lastIOSEnter timing values and making sure to clear the lastIOSEnter fallback timeout

jljorgenson18:
On older iOS devices (like an iPhone 6), the timing between a keydown for pressing enter and when the enter mutation occurs can sometimes be as high as 175ms. With the current lastIOSEnter implementation, you occasionally can get a double enter bug where both the iOS enter fallback executes a custom enter and then immediately after, the original native enter mutates the dom.

Using timeout values of around 200ms seems to handle this problem most of the time, even though that timeout timeline is a bit long. I think ideally we would go with a more deterministic approach on when to use a fallback but I can't think of a way to do that.

### DIFF
--- a/src/domchange.js
+++ b/src/domchange.js
@@ -151,7 +151,7 @@ export function readDOMChange(view, from, to, typeOver, addedNodes) {
   let nextSel
   // If this looks like the effect of pressing Enter (or was recorded
   // as being an iOS enter press), just dispatch an Enter key instead.
-  if (((browser.ios && view.lastIOSEnter > Date.now() - 100 &&
+  if (((browser.ios && view.lastIOSEnter > Date.now() - 225 &&
         (!inlineChange || addedNodes.some(n => n.nodeName == "DIV" || n.nodeName == "P"))) ||
        (!inlineChange && $from.pos < parse.doc.content.size &&
         (nextSel = Selection.findFrom(parse.doc.resolve($from.pos + 1), 1, true)) &&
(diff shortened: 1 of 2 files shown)


jljorgenson18:
Another thing we could look into is if a keyup fires or if intermediate input events fire. The flow where this was very apparent was when input events of type insertReplacementText occurred between keydown and keyup. Maybe we could extend the timing vars if we see input events and maybe we could clear the lastIOSEnter when we see keyup.

jljorgenson18:
I also want to add that this unfortunately could not be reproduced on the iOS simulator. We had to track down an actual iPhone 6 to repro the bug.

marijnh:
> Another thing we could look into is if a keyup fires or if intermediate input events fire

The idea is that enter will usually mutate the DOM, and thus cause `readDOMChange` to be called. That'll then pick up the fact that enter was pressed and handle it as quickly as the machine can manage. Upping the timeout shouldn't degrade experience in typical cases, and should only be noticeable when the native behavior, for whichever reason, doesn't change the DOM. So I think this patch sounds reasonable.

jljorgenson18:
Thank you! And what I meant by using input events and keyup events is that if these long timeouts become a problem, we could use other events as hints that we should either wait longer for the enter event to mutate the dom or we should stop waiting when a key up fires. However as you said, this should be fine and shouldn't really degrade the experience.

PR #76, opened 2020-06-23
Pass plainText to transformPastedText and clipboardTextParser props

pocke:
# Problem

I'd like to use `clipboardTextParser` prop to convert pasted text.
Specifically, I tried to use it to convert pasted markdown text to ProseMirror's node. For example, pasted `> foo` is converted to `blockquote` node.

It almost works well, but I have a problem related to `plainText`.

I expect to disable the converting feature when a user pastes text with `Control + Shift + v`, but `clipboardTextParser` isn't aware of `plainText` or not. So actually it always converts the pasted text to ProseMirror's node.

# Solution

Pass `plainText` to `clipboardTextParser`.
By this change, I can disable the converting feature with Shift key by the following code.

```javascript
new Plugin({
  props: {
    clipboardTextParser(text, $context, plainText) {
      if (plainText) {
        return null;
      }

      return convert(text, $context);
    },
  },
});
```

And I think `transformPastedText` prop has the same problem, so I also made the same change to the prop.

### DIFF
--- a/src/clipboard.js
+++ b/src/clipboard.js
@@ -40,9 +40,9 @@ export function parseFromClipboard(view, text, html, plainText, $context) {
   if (!html && !text) return null
   let asText = text && (plainText || inCode || !html)
   if (asText) {
-    view.someProp("transformPastedText", f => { text = f(text) })
+    view.someProp("transformPastedText", f => { text = f(text, plainText) })
     if (inCode) return new Slice(Fragment.from(view.state.schema.text(text)), 0, 0)
-    let parsed = view.someProp("clipboardTextParser", f => f(text, $context))
+    let parsed = view.someProp("clipboardTextParser", f => f(text, $context, plainText))
     if (parsed) {
       slice = parsed
     } else {
--- a/src/index.js
+++ b/src/index.js
@@ -516,16 +516,19 @@ function needChromeSelectionReset(context, root) {
 //   the clipboard. When not given, the value of the
 //   [`domParser`](#view.EditorProps.domParser) prop is used.
 //
-//   transformPastedText:: ?(text: string, plainText: bool) → string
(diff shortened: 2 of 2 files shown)


marijnh:
This seems reasonable. Could you rename the argument to `plain` (the "text" part seems implicit) in the type signatures and add a note on what it does to the doc comment?

pocke:
> This seems reasonable. Could you rename the argument to `plain` (the "text" part seems implicit) in the type signatures and add a note on what it does to the doc comment?

Thanks for your comment!
I've updated the argument name and the doc. But English is not my first language, so please tell me if the wording is wrong.


Thanks.

marijnh:
On closer look, I think maybe this flag should also be true when text pasting is forced because the selection is in a code block (see the `text && (plainText || inCode || !html)` condition in `parseFromClipboard`). Would that make sense, in your use case?

pocke:
> On closer look, I think maybe this flag should also be true when text pasting is forced because the selection is in a code block (see the `text && (plainText || inCode || !html)` condition in `parseFromClipboard`). Would that make sense, in your use case?

Well, it is reasonable, but I actually only use `clipboardTextParser` and it does not reach `clipboardTextParser` if `inCode` is true. 
Because of https://github.com/ProseMirror/prosemirror-view/blob/593bee904065d78515a57c0ee9da4bfc40cf076e/src/clipboard.js#L44

So `inCode` only affects `transformPastedText`, and I'm not sure the `plain` flag should be true for `transformPastedText`. Currently I don't have any plan to use `transformPastedText` in our app, so I don't have any use case for it. Sorry.

marijnh:
Makes sense. Merged as e3159d9e6d15d82 and touched up in bdb44f83c09545f

PR #78, opened 2020-07-26
update documentation of`nodeViews`

BrianHung:
After updating the type of `node` to include `Mark` and `getPos` to include `bool`, the reference documentation looks like

Screen Shot 2020-07-25 at 9 38 51 PM

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -527,7 +527,7 @@ function changedNodeViews(a, b) {
 //   Can be used to transform pasted content before it is applied to
 //   the document.
 //
-//   nodeViews:: ?Object<(node: Node, view: EditorView, getPos: () → number, decorations: [Decoration]) → NodeView>
+//   nodeViews:: ?Object<(node: union<Node, Mark>, view: EditorView, getPos: union<() → number, bool>, decorations: [Decoration]) → NodeView>
 //   Allows you to pass custom rendering and behavior logic for nodes
 //   and marks. Should map node and mark names to constructor
 //   functions that produce a [`NodeView`](#view.NodeView) object


marijnh:
I think this creates a confusing type. There is never actually a function that accepts these union types — rather, the type of the function is, depending on whether it handles a node or a mark, one of two specific types. But updating the type like that would create an even longer, uglier type. So maybe a better approach is to just keep the type difference in the docs, rather than complicating the documented type?

BrianHung:
> There is never actually a function that accepts these union types — rather, the type of the function is, depending on whether it handles a node or a mark, one of two specific types.

So something more like this?
```
nodeViews:  ?Object<fn(node: Node, view: EditorView, getPos: fn() → number, decorations: [Decoration])→ NodeView> 
| ?Object<fn(mark: Mark, view: EditorView, getPos: bool, decorations: [Decoration]) → MarkView>
```

> So maybe a better approach is to just keep the type difference explanation in the written docs, rather than complicating the documented type?

This works too. My motivation was just to update both function type definitions for both the documentation and at types/prosemirror-view, since I'm moving stuff to typescript. Right now, the definitions at types only account for customNodeViews for nodes and not for marks, which was causing typescript compilation errors.

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b4f12083cb913c0f791b940522be7280b1e391ec/types/prosemirror-view/index.d.ts#L569-L576

marijnh:
I think I prefer updating the type in DefinitelyTyped to be the union-of-function-types while leaving the docs as they are to avoid having an ugly, line-wrapped type.

PR #79, opened 2020-08-09
Remove redundant white-space style

kepta:
It looks like the duplicate white-space is unintentional, so I am sending this PR to remove it.


marijnh:
The duplication is intentional, and provides a fallback for browsers that don't support `break-spaces` (they'll use `pre-wrap`).

PR #80, opened 2020-08-17
Fix incorrect ctrl+drag behavior.

AlansCodeLog:
To fix https://discuss.prosemirror.net/t/weird-ctrl-drag-selection-behavior/3072

Now initially I made a simple fix, just making sure `view.dragging.move` was updated during drag events and on drop as well, to correctly emulate the browser's behavior.

This worked, on the surface, but it made the ctrl+drag copy behavior impossible to stop. I looked into this a bit and found out about [`e.dataTransfer.dropEffect`](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/dropEffect), which should really be what determines `view.dragging.move`. But, there is a [bug in chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=509752&q=39399). It is always `"none"` in chromium based browsers. It can't be changed from dragover/enter listeners like it should be (the last listener should determine what it's set to in the drop event, [example](https://jsfiddle.net/nLwy3e59/), compare with firefox where it works). I played around with different solutions and I think this one is the one has the fewest drawbacks. The solution should be checked in more browsers though since in any non-chromium browsers this would now rely on `e.dataTransfer.dropEffect` functioning correctly.

The other option would be to use the chrome fallback on all browsers so they all behave the same. The only thing this would mean, as far as I can tell, is that you would not be able to set `e.dataTransfer.dropEffect` to `"none"`, but then, if we didn't want the content moved or copied, we would just `e.preventDefault(); return true;` and we were already ignoring the different dropEffects.

The only thing I could not fix is that the cursor does not change to indicate whether the drag modifier is being held. This is because the dragover/dragenter handler's do `e.preventDefault()`, but from what I understand this is recommended and shouldn't be removed. The good news is that it's possible to work around it. Without `e.preventDefault()`, the cursor can't be changed on drag events, but with `e.preventDefault()` they can, so it's just a matter of adding a some move/copy css classes to the editor corresponding to the state of `e.dataTransfer.dropEffect` and also getting the right images since these do not seem to be included in the possible cursor properties. I'll try to look into it properly this week.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -61,6 +61,18 @@ export class EditorView {
     // When editor content is being dragged, this object contains
     // information about the dragged slice and whether it is being
     // copied or moved. At any other time, it is null.
+    // Whether move is true depends on what the `e.transferData.dropEffect`
+    // is set to during the event (`"copy"` will set move to `false`, all other values
+    // will set move to `true`).
+    // Note that ProseMirror will patch `dropEffect` for events it has handled
+    // (e.g. [stopEvent](#view.NodeView.stopEvent) will have the patched value,
+    // but not [handleDOMEvents](#view.EditorProps.handleDOMEvents) which run
+    // before prosemirror handles any events). If it receives `"none"` it will fall back
+    // to checking the platform specific drag modifier keys and set
+    // `e.transferData.dropEffect` accordingly, otherwise it will use the value set
+    // on the event. This means intentionally setting it to `"none"`
+    // (e.g.in a [handleDOMEvents](#view.EditorProps.handleDOMEvents) handler) will
+    // have no effect. To prevent a drop, you can just `e.preventDefault(); return true;`.
     this.dragging = null
 
     initInput(this)
--- a/src/index.js
+++ b/src/index.js
@@ -61,18 +61,6 @@ export class EditorView {
     // When editor content is being dragged, this object contains
     // information about the dragged slice and whether it is being
     // copied or moved. At any other time, it is null.
-    // Whether move is true depends on what the `e.transferData.dropEffect`
-    // is set to during the event (`"copy"` will set move to `false`, all other values
-    // will set move to `true`).
-    // Note that ProseMirror will patch `dropEffect` for events it has handled
-    // (e.g. [stopEvent](#view.NodeView.stopEvent) will have the patched value,
-    // but not [handleDOMEvents](#view.EditorProps.handleDOMEvents) which run
-    // before prosemirror handles any events). If it receives `"none"` it will fall back
-    // to checking the platform specific drag modifier keys and set
-    // `e.transferData.dropEffect` accordingly, otherwise it will use the value set
-    // on the event. This means intentionally setting it to `"none"`
-    // (e.g.in a [handleDOMEvents](#view.EditorProps.handleDOMEvents) handler) will
-    // have no effect. To prevent a drop, you can just `e.preventDefault(); return true;`.
     this.dragging = null
 
     initInput(this)
--- a/src/input.js
+++ b/src/input.js
@@ -552,40 +552,30 @@ editHandlers.paste = (view, e) => {
 class Dragging {
   constructor(slice, move) {
     this.slice = slice
-    this.move = move
+    this._move = move
+    // Ideally view.dragging.move would be set by `e.dataTransfer.dropEffect`
+    // but due to a bug in chrome the fix to make it work is confusing
+    // so for now we ignore it completely, but allow any event handlers with
+    // access to the view to change how that move will be handled by setting `overrideMove`.
+    // Unlike (the correct implementation of) `dropEffect`, `overrideMove` is not
+    // reset/calculated (or in the case of chrome, reset) with every event handler.
+    this.overrideMove = undefined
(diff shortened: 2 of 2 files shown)


marijnh:
> The other option would be to use the chrome fallback on all browsers so they all behave the same.

I'd prefer that—it'd be a bit less complex, and thus less error-prone. Given that this Chrome issue has existed for years, I don't expect to be able to move to the 'proper' solution here anytime soon.

AlansCodeLog:
Okay, I'll change it and document the "limitation" in `view.dragging`. 

What about the cursor issue? Would you be open to the solution of adding a class to the editor depending on the `dropEffect`, I was thinking something like `drop-effect-move/copy`? Would just doing `view.dom.classList.add/remove` from the event handlers be okay?

AlansCodeLog:
I think I've come up with a nice variation of idea 2. `handleDrop` would remain unable to modify `view.dragging`\*, but any nodeview can always use `stopEvent` during drop to still modify view.dragging, and any plugin can still modify `view.dragging` during `handleDOMEvents.drop`. If you like it I'll clean up the commits.

\* I just realized, currently one could always save a reference to `dragging` to then modify during `handleDrop`. By providing a proper mechanism for changing it I can remove this possibility if you want.

marijnh:
> Here's the diff of just the this last solution against master.

Wouldn't this also work without `overrideMove` and `_move`, by just setting `move` (as a regular property) directly?

AlansCodeLog:
No. I mean there's different ways one can approach doing this (like a single variable that is not a boolean), but I was trying to fix it without drastic changes.

The problem is `move` should be set continuously during dragging and *needs* to be set one last time on drop. The drop handler has no way of knowing whether it should use the existing value of `move` (because external code set it) or overwrite it because it doesn't match the state of the drag modifier key. And because drop sets `move` then makes `view.dragging = null` it makes it impossible afaik to override it from external code.

For example, if a plugin/nodeview sets `view.dragging.move = true` during a drop event because they don't want to allow the copy behavior (i.e. my initial problem). The internal drop handler will overwrite it. That's why I created `_move` (internal) and `overrideMove` (external). This has the nice added benefit that one plugin/nodeview can set `overrideMove` and another can set it to false or unset completely if needed and they can do so at any moment when `view.dragging` is defined.

Say one is pressing/un-pressing the drag modifier during the dragging. The behavior would look something like this:

| \_move | overrideMove                                                         | get move\(\) |
|--------|----------------------------------------------------------------------|--------------|
| true   | undefined                                                            | true         |
| false  | undefined                                                            | false        |
| true   | true \(external code prevents copy\)                                 | true         |
| false  | ↓ \(the value "sticks" and is not set with every event like \_move\) | true         |
| true   | ↓                                                                    | true         |
| false  | false \(other external code allows copy\)                            | false        |
| true   | ↓                                                                    | false        |
| false  | undefined \(other external code unsets the override\)                | false        |
| true   | ↓                                                                    | true         |
| false  | ↓                                                                    | false        |

Does that make a bit more sense?

marijnh:
> For example, if a plugin/nodeview sets view.dragging.move = true 

That property isn't even supposed to be mutable to external code, so unless you have a compelling example, I don't really see this as a relevant scenario.

AlansCodeLog:
Sorry, I'm confused, I thought that's what you were asking when you asked "Wouldn't this also work without overrideMove and _move, by just setting move (as a regular property) directly?"

marijnh:
> If we only used one property, move, than how would plugins/nodeviews change what happens on drop?

They can't, unless they use `handleDrop`. This never was a feature that was supposed to be supported, is what I'm saying.

AlansCodeLog:
Right but then the plugin has to rewrite all the insertion code when it's not necessary. 

The way I see it, the ideal solution would be looking at `dropEffect` which would allow changing how drop is handled during dragging, but since patching it is weird and confusing, the next best thing imo is for prosemirror to handle it until that's fixed at least. I don't really see the issue with supporting this. There's plenty of times when one has to use `handleX` to change something while still allowing prosemirror to handle the event in the end so as to not duplicate code, e.g. changing the slice during `handlePaste/Drop`. It seems wrong that one can change the slice and leave prosemirror to handle the insertion but not whether it's copied or moved.

I mean if you're really against it, that's unfortunate, I guess I'll just submit the fix for when `move` is set. There's still the very ugly workaround of keeping a copy of `view.dragging` around to then modify in a `handleDrop` or using my fork.

marijnh:
> It seems wrong that one can change the slice...

That, too, isn't actually something that's supported. As with all objects in the library, if the docs don't tell you you can mutate it, you shouldn't. You can influence the slice, if it comes from an external source, by defining a clipboard parser.

I've pushed c4b1100, which fixes the original issue.

AlansCodeLog:
So... I ended up having to override the paste/drop handlers completely 🤣 because I switched from a nested schema which was insanely difficult to work with to a flat structure with levels (which required completely custom handling but made everything else easier), so I relent.

But I still think the library could make handling cases like this easier. If not by changing how the handlers work or allowing mutability, then providing methods to replicate the default insertion, so one doesn't have to re-duplicate code.

PR #81, opened 2020-09-23
fix: Incorrectly parsing content from the clipboard

lastnigtic:
I‘ve found a case that when dragging out of window and then quickly dragging a node will cause an incorrect parse behavior.
Because when the second drag starts, the previous `dragend` hasn't clean up the `view.dragging`.
It is easier to reproduce from a computer with poor performance.
Below is what I reproduce at [official website](https://prosemirror.net)

drag-issuse-min

### DIFF
--- a/src/input.js
+++ b/src/input.js
@@ -582,7 +582,10 @@ handlers.dragstart = (view, e) => {
 }
 
 handlers.dragend = view => {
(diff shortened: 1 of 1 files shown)


lastnigtic:
I'm not sure if this is the best solution. Please have a check.
Thanks!

marijnh:
Thanks, the fix looks good.

PR #82, opened 2020-09-28
Fix paste when using SHIFT+INS

AndersMad:
Paste keyboard shortcut std. is also **SHIFT + INSERT** - this however will always paste as `plainText` for the `parseFromClipboard` because of (**CTRL + SHIFT + V**). To avoid this I added that the CTRL key also needs to be held while SHIFT is to get a plain text paste.

Not tested on Mac but I assume that the CTRL is Cmd there.

### DIFF
--- a/src/input.js
+++ b/src/input.js
@@ -530,7 +530,7 @@ function capturePaste(view, e) {
 }
 
 function doPaste(view, text, html, e) {
-  let slice = parseFromClipboard(view, text, html, view.shiftKey, view.state.selection.$from)
+  let slice = parseFromClipboard(view, text, html, view.shiftKey && view.ctrlKey, view.state.selection.$from)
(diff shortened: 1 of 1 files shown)


marijnh:
`view.ctrlKey` isn't a thing that is defined, so that will always be false, and this just breaks plain text pasting.

I'm not interested in changing the way plain text paste is done. A solution that detects shift-insert and disables the plain-text paste behavior for that would be welcome.

AndersMad:
I'm sry, I just assumed that it was there + all 3 ways worked when testing: `CTRL + INS`, `CTRL + SHIFT + INS` (as text) and `SHIFT + INS` (not as plain text after this) - and that's because that (Chrome) will only have text/plain clipboard data on CTRL + SHIFT + INS 🤦‍♂️ .. I'll get back to the SHIFT key thing later then - in case the paste source comes from other source..

PR #83, opened 2020-12-10
Checking if iPadOS using maxTouchPoints so iOS workarounds work for iPadOS and iOS

jljorgenson18:


jljorgenson18:
#https://github.com/ProseMirror/prosemirror/issues/1111

willdarwin:
It gets wrong result cause some **Windows devices** also has maxTouchPoints value. See pic below
image

marijnh:
That looks like an Android device -- are those setting `navigator.vendor` to "Apple Computers"? Because this test is only done for browsers that are detected as Safari.

willdarwin:
> That looks like an Android device -- are those setting `navigator.vendor` to "Apple Computers"? Because this test is only done for browsers that are detected as Safari.

You are right, as the latest code checks navigator.vendor first, it works fine now. 

I also checked Safari/Chrome/QQ browser in iOS, the naviagator.vendor value are all "Apple Computers".

The Pic is from a touchable Windows computer: the platform is 'Win32', the vendor is 'Google Inc.'

PR #84, opened 2021-01-10
Enable a image in PNG format to be pasted into a Prosemirror document.

ronkok:
This PR enables Prosemirror to paste a image in the PNG format from the system clipboard into a Prosemirror document. The resulting document will contain an `` element whose `src` is a URI base 64 conversion of the PNG.

The PR's code is simpler than one might expect.  No code is necessary to wrap that PNG in a ``. The browser does that itself any time a PNG is pasted into a `contenteditable` element. All this PR does is call `capturePaste()` when appropriate.

I find that `capturePaste()` will properly append the `` to the Prosemirror document and that the Prosemirror undo stack continues to operate properly.

The motivation for this PR is to enable authors to copy screenshots from the system clipboard into a Prosemirror document as discussed in https://discuss.prosemirror.net/t/paste-systems-screenshot/2402/3.

### DIFF
--- a/src/input.js
+++ b/src/input.js
@@ -542,7 +542,9 @@ function doPaste(view, text, html, e) {
 editHandlers.paste = (view, e) => {
   let data = brokenClipboardAPI ? null : e.clipboardData
   let html = data && data.getData("text/html"), text = data && data.getData("text/plain")
-  if (data && (html || text || data.files.length)) {
+  if (data.items && data.items.length > 0 && data.items[0].type === "image/png") {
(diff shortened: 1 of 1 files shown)


marijnh:
Attached patch does something similar in a more general way. Could you see if it works for you?

ronkok:
Yes, your patch works better than what I wrote. Go with it. I'm looking forward to having the ability to paste screenshots.

ronkok:
Closing this PR in favor of commit a67f66d.

PR #86, opened 2021-02-22
trim className before split to avoid empty className

1inus:
in `function patchAttributes(dom, prev, cur)`, className like 'someClassName ' get an array contains empty element after split(' ') called. `dom.classList.add(emptyString)` will case an exception : `Uncaught (in promise) DOMException: Failed to execute 'add' on 'DOMTokenList': The token provided must not be empty.`

### DIFF
--- a/src/viewdesc.js
+++ b/src/viewdesc.js
@@ -988,7 +988,7 @@ function patchAttributes(dom, prev, cur) {
       dom.setAttribute(name, cur[name])
   if (prev.class != cur.class) {
     let prevList = prev.class ? prev.class.split(" ") : nothing
-    let curList = cur.class ? cur.class.split(" ") : nothing
+    let curList = cur.class ? cur.class.trim().split(" ") : nothing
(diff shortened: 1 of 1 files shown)


marijnh:
We're targeting platforms (specically IE11) that don't have `trim`. Maybe express this as a `filter` call after the split instead to avoid that?

BrianHung:
[MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim) and [caniuse](https://caniuse.com/mdn-javascript_builtins_string_trim) says that `String.prototype.trim` has been supported since IE10. 

Using `.split(" ").filter(Boolean)` is more compatible (with up to [IE9](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter)) but [prosemirror homepage](https://prosemirror.net/) says compatibility starts at IE11.

marijnh:
Ah, I guess it's `trimStart`/`trimEnd` that aren't in IE11. Thanks for pointing that out. I agree the filter patch is nicer.

PR #87, opened 2021-03-03
Pass inner decorations to nodeView on init and update

jonathonherbert:
This PR passes the decorations applied to a nodeView's content to its initialiser and update() method.

This enables the use case discussed [here](https://discuss.prosemirror.net/t/is-it-possible-to-get-hold-of-decorations-that-cover-the-range-managed-by-a-nodeview-in-that-nodeview/3544) – Prosemirror instances nested within nodeViews will be able to make use of their parent's decorations.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -540,7 +540,7 @@ function changedNodeViews(a, b) {
 //   Can be used to transform pasted content before it is applied to
 //   the document.
 //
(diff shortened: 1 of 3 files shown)


marijnh:
Merged, and followed up by https://github.com/ProseMirror/prosemirror-view/commit/25d0c11a49ece78d1288ac5e7d07e1434177bfde

PR #88, opened 2021-03-07
Fix scroll when editor has css transform:scale()

rbasto1:
This fixes an issue with editors inside elements with style "transform: scale(xx)" that sometimes causes scrolling to take place too early (when scale > 1) or after the cursor already left the window boundary (when scale < 1), since clientWidth and clientHeight being used to offset the scrollbars are unscaled.

### DIFF
--- a/src/domcoords.js
+++ b/src/domcoords.js
@@ -12,9 +12,11 @@ function getSide(value, side) {
 
 function clientRect(node) {
   let rect = node.getBoundingClientRect()
+  // Adjust for elements with style "transform: scale()"
+  let scale = (rect.width / node.offsetWidth) || 1;
   // Make sure scrollbar width isn't included in the rectangle
-  return {left: rect.left, right: rect.left + node.clientWidth,
-          top: rect.top, bottom: rect.top + node.clientHeight}
+  return {left: rect.left, right: rect.left + node.clientWidth * scale,
+          top: rect.top, bottom: rect.top + node.clientHeight * scale}
 }
 
 export function scrollRectIntoView(view, rect, startDOM) {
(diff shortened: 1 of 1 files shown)


marijnh:
This seems very limited—it'll only work for scale, not other tranformations, and even there only for transformations relative to the top-left corner of the viewport. That may work for your case but doesn't seem like a reasonable thing for the library to do.

How about, instead, we introduce a prop that provides two-way coordinate tranformation functions, and call those at the appropriate points? That way, we could offload the complexity to client code, which may know more about the transformation (such as in your case).

rbasto1:
This actually works for scale() from any relative position, since `left` and `top` are coming from `getBoundingClientRect` and are therefore already scaled. Other transformations like translate need no extra code, they are already working for the same reason. It is only scale() that is giving me issues.
You do have a point: this assumes a scale that is symmetric for x and y. I could update the code to calculate scaleX and scaleY separately.
Why would the understanding of a built-in, vanilla css attribute be unreasonable to the library to be aware of? I don't think it's the responsibility of the client to help the library understand how to calculate properly the coordinates of the dom. The current code is plainly incorrect: it assumes you can mix `clientWidth` and `clientHeight` with the rect from `getBoundingClientRect`, but these are not the same units unless in the special case of a scale of 1.0.

marijnh:
I initially read this as a way to solve transformation handling in general (issues like code positioning things inside the transformed element based on DOM measurements, such as the drop cursor does). That's a problem that browsers seemed to be interested in solving [at one point](https://hacks.mozilla.org/2014/04/coordinate-conversion-made-easy/), but they never actually implemented the proposed solution. And it seems a library, without such browser support, cannot do a proper job of this, so I figured the least confusing thing is to do nothing at all then.

But I realize this is actually smaller in scope, just avoiding a coordinate system mismatch in that specific computation. That's a lot more reasonable.

marijnh:
(I still want to point out that this will help exactly nothing when rotation or shearing transformations are applied, but since the editor falls apart anyway in such a scenario, that's not really a problem.)

PR #89, opened 2021-03-15
trim both prev class and cur class before splitting #86

BrianHung:
See https://github.com/ProseMirror/prosemirror-view/pull/86


BrianHung:
trim only removes leading and trailing white space; boolean filter removes unnecessary white space inside the string as well

Example:
```js
" a  b  c ".trim().split(" ") // ["a", "", "b", "", "c"]
" a  b  c ".split("").filter(Boolean) // ["a", "b", "c"]
```

PR #91, opened 2021-05-08
Update index.js

4-1-1:
able to get new state after dispatch

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -371,8 +371,10 @@ export class EditorView {
   // easily passed around.
   dispatch(tr) {
     let dispatchTransaction = this._props.dispatchTransaction
-    if (dispatchTransaction) dispatchTransaction.call(this, tr)
-    else this.updateState(this.state.apply(tr))
+    if (dispatchTransaction) {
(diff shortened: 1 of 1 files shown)


marijnh:
Doesn't reading `view.state` after calling dispatch already provide this?

4-1-1:
Any idea to make it without reading `view.state` ?

jljorgenson18:
@4-1-1 Why not just do

```js
          const to =
            $to.pos - $to.parentOffset + path2[path1.length - 3].content.size;
          tr.setSelection(TextSelection.create(state.doc, from, to);
          const content = tr.selection.content();
          const stringContent = stringifyContent(
            content.content,
            isInCode(tr) // Assuming this could take a transaction or a selection object instead
          );
          const node = schema.nodes[type].create(
            {},
            schema.text(stringContent)
          );
          dispatch(tr.replaceSelectionWith(node));
```

marijnh:
Or, if this is a command, read the third argument (the view).

4-1-1:
thanks a lot. It seems like a better way. 
and I close it. @jljorgenson18 @marijnh

PR #92, opened 2021-05-12
check if focusNode of root selection is contained within view dom

BrianHung:
I'm on `prosemirror-view@1.18.4`, and am running into a runtime problem when holding the up or down arrow keys when the prosemirror-gapcursor is at the edge of the document.

The error I'm encountering comes from

```js
function isBlockNode(dom) {
  var desc = dom.pmViewDesc; // TypeError: Cannot read property 'pmViewDesc' of null
  return desc && desc.node && desc.node.isBlock
}
```

with the stack trace being

```js
Uncaught TypeError: Cannot read property 'pmViewDesc' of null
    at isBlockNode (index.es.js:2345)
    at skipIgnoredNodesRight (index.es.js:2321)
    at captureKeyDown (index.es.js:2466)
    at push../node_modules/prosemirror-view/dist/index.es.js.editHandlers.keydown (index.es.js:3374)
    at HTMLDivElement.view.dom.addEventListener.view.eventHandlers.<computed> (index.es.js:3299)
```

If we open up `skipIgnoredNodesRight`, the issue cascades like so:

```js
function skipIgnoredNodesRight(view) {
  var sel = view.root.getSelection();
  var node = sel.focusNode, offset = sel.focusOffset; // 1. node is not contained within view.dom, but is not null
  if (!node) { return }
  var len = nodeLen(node);
  var moveNode, moveOffset;
  for (;;) {
    if (offset < len) {
      if (node.nodeType != 1) { break }
      var after = node.childNodes[offset];
      if (isIgnorable(after)) {
        moveNode = node;
        moveOffset = ++offset;
      }
      else { break }
    } else if (isBlockNode(node)) { // 3. because node is now null, TypeError
      break
    } else {
      var next = node.nextSibling;
      while (next && isIgnorable(next)) {
        moveNode = next.parentNode;
        moveOffset = domIndex(next) + 1;
        next = next.nextSibling;
      }
      if (!next) {
        node = node.parentNode; // 2. keeps climbing until node is #document, which then node = node.parentNode = null
        if (node == view.dom) { break }
        offset = len = 0;
      } else {
        node = next;
        offset = 0;
        len = nodeLen(node);
      }
    }
  }
  if (moveNode) { setSelFocus(view, sel, moveNode, moveOffset); }
}
```

In my specific case, the `focusNode` of `view.root.getSelection()`  when arrowing down somehow landed here
```html
    <next-route-announcer>
      /* sel.focusNode */
      <p aria-live="assertive" id="__next-route-announcer__" role="alert" style="border: 0px; clip: rect(0px, 0px, 0px, 0px); height: 1px; margin: -1px; overflow: hidden; padding: 0px; position: absolute; width: 1px; white-space: nowrap; overflow-wrap: normal;"></p>
    </next-route-announcer>
  </body>
</html>
```

prosemirror-view

Seems sort of related to #588 and #828 with issues with selections outside of the editor view, and `skipIgnoredNodesLeft` and `skipIgnoredNodesRight` (?). I'm not exactly sure of the interactions between Chrome and prosemirror-gapcursor which lead to the `var sel = view.root.getSelection();` being placed _outside_ of the editor even though `document.activeElement = view.dom` still.

### DIFF
--- a/src/capturekeys.js
+++ b/src/capturekeys.js
@@ -1,7 +1,7 @@
 import {Selection, NodeSelection, TextSelection, AllSelection} from "prosemirror-state"
 import browser from "./browser"
 import {domIndex, selectionCollapsed} from "./dom"
-import {selectionToDOM} from "./selection"
+import {selectionToDOM, hasFocusAndSelection} from "./selection"
 
 function moveSelectionBlock(state, dir) {
   let {$anchor, $head} = state.selection
@@ -61,6 +61,7 @@ function isIgnorable(dom) {
 // Make sure the cursor isn't directly after one or more ignored
 // nodes, which will confuse the browser's cursor motion logic.
 function skipIgnoredNodesLeft(view) {
+  if (!hasFocusAndSelection(view)) return
   let sel = view.root.getSelection()
   let node = sel.focusNode, offset = sel.focusOffset
   if (!node) return
@@ -109,6 +110,7 @@ function skipIgnoredNodesLeft(view) {
 // Make sure the cursor isn't directly before one or more ignored
 // nodes.
 function skipIgnoredNodesRight(view) {
+  if (!hasFocusAndSelection(view)) return
   let sel = view.root.getSelection()
   let node = sel.focusNode, offset = sel.focusOffset
   if (!node) return


BrianHung:
> That's odd—those should only run when the editor has keyboard focus, and I'd assume the browser to move focus out of the editor when the selection leaves it. Can you confirm that `document.activeElement` is still the editor when this occurs?

activeElement

Yep, document activeElement is still the editor.

BrianHung:
> the page itself shifts all the way down, almost as if the arrow keydown event wasn't captured completely

I tried tested whether `if (!hasFocusAndSelection(view)) return` also prevents the page itself from shifting down. It doesn't, which might imply the bug and fix has to be somewhere else..

BrianHung:
> I wonder if there's a way to prevent default when prosemirror gapcursor selection is at the end of a document, while also allowing other types of selections to be natively handled by the browser (which is done when return value of `skipIgnoredNodesRight` is null?).

Hmm, as the previous images suggest, only the gapcursor selection seems to cause a selection where the focusNode is the prosemirror contenteditable dom itself.
```js
var sel = view.root.getSelection();
var node = sel.focusNode, offset = sel.focusOffset; // node == view.dom
```

So returning
```js
function skipIgnoredNodesRight(view) {
  ...
  return view.dom == node
```
might work.

marijnh:
The idea with capturing keys is to only preventDefault stuff that is explicitly being handled by the library. Using that to try and suppress random inappropriate native browser behavior seems very kludgey.

I tried to reproduce this problem with a gapcursor below a `<hr>` in the demo, but couldn't. Have you tried reducing it to its minimal ingredients?

BrianHung:
@marijnh I've noticed the bug is now reproducible on https://prosemirror.net/. Try it out there with a gapcursor below a hr.

marijnh:
> I've noticed the bug is now reproducible on https://prosemirror.net/.

Could you describe more precisely what you're doing? What is the document like, what keys are you pressing/holding, and what does the unexpected behavior look like? I tried making the document a single horizontal rule and holding the vertical arrow keys, and was looking for an error in the console, but I'm not seeing any.

BrianHung:
CleanShot 2021-12-20 at 10 05 51

Sorry, didn't mean bug was fully reproducible with an error existing in the console. But the arrow keydown with gapcursor resulting in viewport scroll away from the text editor itself.

marijnh:
Oh, right, I see. Chrome will often do bizarre things when the cursor isn't in a plain text position. I'm not sure what would be a reasonable strategy for interfering with that, though—the general strategy of the library is to not interfere with arrow keys unless it has a specific action it wants to take for them, and just let the native behavior happen otherwise. In this case, moving down at the end of the document, there is nothing reasonable the editor can do with the key event, and handling it there might interfere with integrations that want to do something special for arrow presses pointing beyond the document end.

PR #93, opened 2021-05-18
Add necessary devDependencies for script test-server

MagnetoLee:
Add necessary devDependencies for script test-server

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,9 @@
     "moduleserve": "^0.7.0",
     "prosemirror-test-builder": "^1.0.0",
     "rollup": "^2.26.3",
-    "@rollup/plugin-buble": "^0.21.3"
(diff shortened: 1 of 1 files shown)


marijnh:
Where are these used in the tests?

MagnetoLee:
After clone and npm install, when npm run test-server will get this

image

I'm not sure what happened but can fixed by add devDependencies.

MagnetoLee:
After clone and npm install, when npm run test-server will get this

image

I'm not sure what happened but can fixed by add devDependencies.

marijnh:
That appears to be going through prosemirror-test-builder, which has these packages specified as peer dependencies. Which version of npm are you using to install the packages?

MagnetoLee:
6.4.1
I will close it, thanks.

PR #96, opened 2021-07-21
 prosemirror-* packages as peer dependencies

kepta:
This change proposes moving the prosemirror-* packages to peer dependency which should give downstream consumers more flexibility and also [npm v7](https://github.blog/2020-10-13-presenting-v7-0-0-of-the-npm-cli/) now supports automatically installing peer dependencies.


marijnh:
Would this not mess things up for people stuck on older versions of NPM?

PR #97, opened 2021-07-24
Update compositionEndedAt when composition cleared manually

KentoMoriwaki:
https://github.com/ProseMirror/prosemirror/issues/1190

### DIFF
--- a/src/input.js
+++ b/src/input.js
@@ -464,7 +464,7 @@ export function clearComposition(view) {
   if (view.composing) {
     view.composing = false
     let event = browser.ie ? document.createEvent("CompositionEvent") : new CompositionEvent("compositionend")
-    view.compositionEndedAt = event.timeStamp;
+    view.compositionEndedAt = event.timeStamp
   }
   while (view.compositionNodes.length > 0) view.compositionNodes.pop().markParentsDirty()
 }
--- a/src/input.js
+++ b/src/input.js
@@ -469,13 +469,9 @@ export function clearComposition(view) {
 }
 
 function timestampFromCustomEvent() {
-  try {
-    return new CompositionEvent("compositionend").timeStamp
-  } catch (e) {
-    let event = document.createEvent("CompositionEvent", {})
-    event.initCompositionEvent("compoositionend", true, true, window, {}, null)
-    return event.timeStamp
-  }
+  let event = document.createEvent("Event")
+  event.initEvent("event", true, true)
+  return event.timeStamp
 }
 
 export function endComposition(view, forceUpdate) {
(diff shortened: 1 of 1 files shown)


marijnh:
The assumption in the current code is that the browser will still fire a compositionend event after we force an end to it like this. Are you running into situations where that doesn't happen?

KentoMoriwaki:
@marijnh Thanks for reviewing!

A compositionend event will be fired, but it does not update `view.compositionEndedAt` because `view.composing` already set to false in the `clearComposition`.

https://github.com/ProseMirror/prosemirror-view/blob/63aebf43770480659ab4a84b406263548f30dc42/src/input.js#L450-L456

Another solution would be to eliminate the `if (view.composing)` condition in the compositionend handler. Would that be better?

marijnh:
Ah, right, good point, that makes sense. Always setting `composition` and `compositionEndedAt` together as in your patch seems reasonable.

A remaining issue is that IE doesn't have `CustomEvent`. Can you replace that with a `document.createEvent`/`event.initEvent`-based solution? (Or maybe there's another simpler way to get an event timestamp, I'm not sure.)

KentoMoriwaki:
I updated to handle cross browser. CompositionEvent constructor is supported by IE. https://developer.mozilla.org/en-US/docs/Web/API/CompositionEvent/CompositionEvent#browser_compatibility

marijnh:
Are you sure that's not just a mistake in MDN? Because the other events (see for example [KeyboardEvent](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/KeyboardEvent#browser_compatibility) explicitly mention that the constructor form isn't supported by IE (thus the need for `createEvent`+`initEvent`).

KentoMoriwaki:
You're right. I tested on IE and fixed it. (we don't need `initEvent` to get timeStamp)

marijnh:
Hm, sorry to continue nitpicking, but (pre-Webkit) Edge browsers will also have `browser.ie` set, and may have adjusted to what other browsers are doing far enough to _not_ have `createEvent` set the timestamp (`createEvent` exists on modern browsers as well, but doesn't set a timestamp yet, leaving that to `initEvent`). I'd prefer a single path for all browsers that just calls both `createEvent` and `initEvent` to be sure.

KentoMoriwaki:
Sorry for the late reply. I updated. How about this?

marijnh:
> I'd prefer a single path for all browsers that just calls both createEvent and initEvent to be sure.

I still do. Those functions are available in all browsers, and also used in the keyboard-event-faking logic for Android, so just use them directly.

marijnh:
Thanks!

PR #98, opened 2021-08-05
DecorationSet: Filter by predicate with range arguments excluded

georgemandis:
Reading the documentation for `DecorationSet` I interpreted the range arguments as optional while still being able to pass-in a predicate like this:

```js
decoration.find((deco) => ...)
``

In practice I have to explicitly pass `undefined` as start and end values:

```js
decoration.find(undefined, undefined, (deco) => ...)
```

This PR updates the method to function the way I expected, but I recognize that might just be syntactic sugar, a little magical and perhaps unwanted. 

I updated the method to recognize when I'm only passing in a single function as an argument, assume that's my predicate and treat the start/end values as `null` so it's applied to the entire range.

### DIFF
--- a/src/decoration.js
+++ b/src/decoration.js
@@ -243,6 +243,10 @@ export class DecorationSet {
   // assumed to match.
   find(start, end, predicate) {
     let result = []
+    if (typeof start === "function" && end == null && predicate == null) {
+      predicate = start
+      start = null
+    }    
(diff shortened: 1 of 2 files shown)


marijnh:
> I interpreted the range arguments as optional while still being able to pass-in a predicate like this:

That's not how optional arguments work in JavaScript, though. The documentation system doesn't really have support for overloads like this, and it's generally a good idea to pass in a range for efficiency, so I prefer not to merge this.

PR #99, opened 2021-08-06
Add .map to DecorationSource interface 

jonathonherbert:
... and implement for DecorationGroup.

## Why?

We're working on a [library](https://github.com/guardian/prosemirror-elements) that aims it easier to model arbitrary combinations of nested fields representing different sorts of data within a Prosemirror instance. This includes nested rich text fields.

It's useful to be able to pass decorations from the parent editor to its children. This lets us represent spellcheck annotations, or other users' cursors, with a single plugin in the parent editor. This was implemented in #87.

In our plugin, when we receive a `DecorationSource` in a child editor, it represents the decorations for a NodeView that contains the entire `Element`, within which that child is one of what may be many nodes. Thus, we must map these decorations through the offset of that node relative to its parent.

This PR adds a `.map` method to `DecorationSource`, and implements it for `DecorationGroup`, to address that use case. Here's an example of it working – the parent editor is using [prosemirror-typerighter](https://github.com/guardian/prosemirror-typerighter), and the child node in a prosemirror-elements `Element` is displaying decorations correctly.

Screenshot 2021-08-06 at 09 39 15

## Questions

I'm not quite sure of the best way to write a unit test for this new method – any guidance v. welcome.

### DIFF
--- a/src/decoration.js
+++ b/src/decoration.js
@@ -419,7 +419,7 @@ export class DecorationSet {
 // decorations. Implemented by [`DecorationSet`](#view.DecorationSet),
 // and passed to [node views](#view.EditorProps.nodeViews).
 //
-// map:: (Mapping, Node) → DecorationSet
+// map:: (Mapping, Node) → DecorationSource
 // Map the set of decorations in response to a change in the
 // document.
 
@@ -440,11 +440,10 @@ class DecorationGroup {
   }
 
   map(mapping, doc) {
-    const mappedMembers = [];
-    for (let i = 0; i < this.members.length; i++) {
-      mappedMembers.push(this.members[i].map(mapping, doc, options))
-    }
-    return DecorationGroup.from(mappedMembers)
+    const mappedDecos = this.members.map(
+      member => member.map(mapping, doc, noSpec)
+    )
+    return DecorationGroup.from(mappedDecos)
   }
 
   forChild(offset, child) {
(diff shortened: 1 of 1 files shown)


marijnh:
That's an interesting, if obscure, way to use these.

Why not just use `this.members.map(m => m.map(mapping, doc, options))` in the implementation?

And I think you can leave off the options argument in the `DecorationSource` interface, since `onRemove` isn't going to be meaningful for an opaque group of decorations, and it helps keep the docs simple (and without duplication).

jonathonherbert:
> That's an interesting, if obscure, way to use these.

Haha, yes – but quite handy! There are a few instances where this is useful. For example, we can represent the cursors in a collaborative editing context in the same way, with one parent plugin producing decorations which are then rendered in child editors. As an example, here's three editors side by side connected by collab, displaying cursors in nested fields.

Screenshot 2021-08-12 at 09 17 08

> Why not just use `this.members.map(m => m.map(mapping, doc, options))` in the implementation?

This would work – but it's not part of a documented interface, so I assumed `members` was not a public API. My reasoning is that extending the `DecorationSource` interface allows us to use `.map` without caring about the underlying type `DecorationGroup | DecorationSet`, which IIUC is the reason there's a common interface.

> And I think you can leave off the options argument in the DecorationSource interface, since onRemove isn't going to be meaningful for an opaque group of decorations, and it helps keep the docs simple (and without duplication).

Thanks for the feedback, I've just done that.

marijnh:
> so I assumed members was not a public API.

No, I mean in the PR, that could make the new `map` method a lot simpler.

marijnh:
Oh also the type comment for `map` should return a `DecorationSource` I believe.

marijnh:
I do write manual loops whenever I'm worried about performance, but this isn't going to be critical path in any conceivable scenario, so optimizing for terseness seems preferable.

Looks good now, merging!

PR #102, opened 2021-08-13
Fix of Backspace/Delete not working for RTL characters in shadowDOM

maksim-shlapak:
Fix ProseMirror/prosemirror#1195
In case when editor inside ShadowDOM and RTL characters was used Backspace/Delete works incorrect. This happened because global function getSelection causes incorrect result and endOfTextblockHorizontal return incorrect value


marijnh:
Thanks for tracking that down!

PR #103, opened 2021-09-06
Use `::` to signify pseudo element

amk221:
Not strictly necessary, but adds clarity imo

### DIFF
--- a/style/prosemirror.css
+++ b/style/prosemirror.css
@@ -33,7 +33,7 @@ li.ProseMirror-selectednode {
   outline: none;
 }
 
-li.ProseMirror-selectednode:after {
+li.ProseMirror-selectednode::after {
   content: "";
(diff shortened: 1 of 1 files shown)


marijnh:
That's not a style rule this project uses, so I'm passing on this PR.

amk221:
No worries, thanks for reviewing!

Can you elaborare on 
> That's not a style rule this project use

marijnh:
The preference for `::` is not something I'm doing in the css files for ProseMirror. I also wouldn't accept a PR adding semicolons to the JavaScript files or similar.

amk221:
Alrighty 🙂

Just to make sure we're on the same page:

From MDN
> As a rule, double colons (::) should be used instead of a single colon (:). This distinguishes pseudo-classes from pseudo-elements

PR #105, opened 2021-09-13
Add a class name to distinguish ProseMirror hack nodes from regular nodes

amk221:
ProseMirror looks at the schema to decide when to insert hack nodes, this is fine most of the time. But an `inline` node may have styles that make it display `block`, contrary to its schema spec. Meaning the hack is applied when it is not needed, which can lead to display issues.

This PR adds a class name to allow hack nodes to be differentiated from nodes that are actually part of the document. It does not fix the above, but it goes some way to allow developers to work around the issue if it affects them.

### DIFF
--- a/src/viewdesc.js
+++ b/src/viewdesc.js
@@ -1258,6 +1258,7 @@ class ViewTreeUpdater {
     } else {
       let dom = document.createElement(nodeName)
       if (nodeName == "IMG") dom.className = "ProseMirror-separator"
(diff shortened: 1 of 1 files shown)


marijnh:
> But an inline node may have styles that make it display block, contrary to its schema spec. 

(For those running across this: No it may not. That is not something the library supports. Don't do that.)

amk221:
Apologies, by "may have", I meant like it could, possibly have a `style` attribute (if it were pasted by a user for example). 
I didn't mean it is allowed or endorsed.

PR #106, opened 2021-09-14
Fix matchesNode of null runtime error

flaviouk:
When an editor is removed from the page, it seems that `matchesNode` gets hit which means that if this.docView is null, it causes a runtime error.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -146,7 +146,7 @@ export class EditorView {
 
     let scroll = reconfigured ? "reset"
         : state.scrollToSelection > prev.scrollToSelection ? "to selection" : "preserve"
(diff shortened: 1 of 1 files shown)


marijnh:
Do you have a stacktrace for whatever is calling `updateState` after the editor has been removed from the document? This doesn't normally happen.

flaviouk:
Does this help @marijnh ?
image

marijnh:
Not a lot. It suggests that a backspace handler is involved, but why would that be firing when the editor isn't in the DOM? Could it be that you're removing the editor via an event and not stopping further handling of that event?

sharno:
This seems to be coming from ueberdosis/tiptap#2380

marijnh:
Thanks for the link. Closing this issue here.

PR #107, opened 2021-09-22
Fix collapsing whitespace in Firefox

ausi:
I debugged https://github.com/ProseMirror/prosemirror/issues/651 and found out that Gecko handles trailing spaces differently from Blink/WebKit by using a trailing `<br>` element instead of a `&nbsp;`.

In this pull request I added a check for this case to the textblock hacks so that we get a `<br class="ProseMirror-trailingBreak">` element if the last text node ends in a whitespace character.

All issues described in https://github.com/ProseMirror/prosemirror/issues/651 no longer appear after this change. (tested with `white-space: normal`)

It might be possible to generally switch to `white-space: normal` now to fix https://github.com/ProseMirror/prosemirror/issues/857


ausi:
> `white-space: normal` will make it impossible to have multiple spaces in a row in your content, no?

Multiple spaces still work. In a series of multiple spaces all spaces but the last one get converted to a non-breaking-space (`U+00A0`).
This should be exactly what you want if you are using `white-space: normal` I think.

> so I am not convinced this solves that issue.

It definitely fixes the issues from ProseMirror/prosemirror#651

If switching the default to `white-space: normal` is a good idea apart from that, might be dependent on many other edge cases I think.

But it should no longer be necessary to warn against using `white-space: normal` after this fix.

marijnh:
>  In a series of multiple spaces all spaces but the last one get converted to a non-breaking-space

Not by ProseMirror—that'll simply render the spaces.

ausi:
> > In a series of multiple spaces all spaces but the last one get converted to a non-breaking-space
> 
> Not by ProseMirror—that'll simply render the spaces.

This is done by the browsers themself in the `contenteditable` mode.

I think this is the specification most browsers have currently implemented: <https://w3c.github.io/editing/docs/execCommand/#canonical-space-sequences>

marijnh:
> This is done by the browsers themself in the contenteditable mode.

Only when you edit a given text node, though.

ausi:
> Only when you edit a given text node, though.

I see. Just tested this by copying text with multiple spaces from a text editor and indeed the spaces get collapsed when pasting into the editor with `white-space: normal`.

For my use case this is perfectly fine, but I see how this could cause issues for other cases.

So I guess `pre-wrap`/`break-spaces` should stay as the default.

Do you agree that the warning can be removed as this pull request currently does? Or should I revert that part?

marijnh:
No, definitely keep the warning. As things are, using collapsing whitespace is not supported with ProseMirror.

ausi:
> No, definitely keep the warning.

Done in d8c4bb1c238d814359a345e576fdfcaaa17aabd2

ausi:
> Is this still something that makes sense to merge?

For me yes ☺️
Merging this would be great and fix https://github.com/ProseMirror/prosemirror/issues/651

ausi:
Should I resolve the conflicts and/or (try to) add a test?

marijnh:
Ah, I just realized this is going to constantly create unnecessary and possibly disruptive `<br>`s when typing at the end of a textblock with the proper supported `white-space` setting. Could you maybe add some flag, managed by the `checkCSS` function in domobserver, that selectively turns this on only on editors that have an unsupported `white-space` styling?

ausi:
> Could you maybe add some flag, managed by the `checkCSS` function in domobserver, that selectively turns this on only on editors that have an unsupported `white-space` styling?

Yes. I will try that.

ausi:
Done in afa44e1afe87abf542cb87ad68ced9f111f688a2. I used a `WeakMap` to ensure that reading the styles is only done once for every EditorView instance.

ausi:
@marijnh anything else missing in this pull request?

marijnh:
Sorry, lost sight of this. Looks good now, except:

 - Could you make it so that it only spits out the warning once, not once for every editor view?
 - Could you add a comment linking https://github.com/ProseMirror/prosemirror/issues/651 somewhere for people wondering what this hack is about?

ausi:
> Could you make it so that it only spits out the warning once, not once for every editor view?

Done in cdfe56b5fbef00352a00bd724d91b35f217e8a25

> Could you add a comment linking https://github.com/ProseMirror/prosemirror/issues/651 somewhere for people wondering what this hack is about?

Done in 38b4a3c63d6fd92dc4bf4766ca1f9a10975eeea7

marijnh:
Thanks!

PR #110, opened 2021-11-05
Fix Markdown syntax in DirectEditorProps section

command-tab:
This PR fixes Markdown link syntax in the prosemirror-view "DirectEditorProps" section under the "plugins" heading. It was previously broken and exposing the Markdown syntax on the rendered version:

image

Thank you for maintaining this project ❤️


marijnh:
Thanks for fixing that!

PR #111, opened 2021-11-07
feat (Widget): Add a destroy function to Decoration.widget

jamesopti:
Adds a `destroy` method to widget decorations per discussion here:

https://discuss.prosemirror.net/t/unmounting-a-decoration-widget-created-with-react/3050/6

### DIFF
--- a/src/decoration.js
+++ b/src/decoration.js
@@ -14,7 +14,11 @@ class WidgetType {
 
   map(mapping, span, offset, oldOffset) {
     let {pos, deleted} = mapping.mapResult(span.from + oldOffset, this.side < 0 ? -1 : 1)
-    return deleted ? null : new Decoration(pos - offset, pos - offset, this)
+    if (deleted) {
+      this.destroy()
+      return null
+    }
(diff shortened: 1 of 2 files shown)


src/decoration.js
```diff
@@ -14,7 +14,11 @@ class WidgetType {
 
   map(mapping, span, offset, oldOffset) {
     let {pos, deleted} = mapping.mapResult(span.from + oldOffset, this.side < 0 ? -1 : 1)
-    return deleted ? null : new Decoration(pos - offset, pos - offset, this)
+    if (deleted) {
+      this.destroy()
+      return null
+    }
```
RatoX, line 20: is this the right moment to call the `destroy`? The DOM element wasn't removed yet.

jamesopti: Ah good call. I realized this should be invoked by implementing `destroy` in `WidgetViewDesc`

Updated.

marijnh:
Thanks! I've followed up with 05c05d9, since I can imagine the destroy function needing access to the actual DOM node that was created for the widget.

PR #112, opened 2021-11-08
remove class attribute if classList is empty

BrianHung:
Not really a bug, but when all the class names of `prev` are removed and class names of `curr` is empty, there is an empty class attribute on the resulting outputted DOM.

https://github.com/ueberdosis/tiptap/issues/2126


marijnh:
Seems reasonable.

PR #113, opened 2021-11-12
Increase bounds check for children in reconcilliation back to 5

seanchambo:
A potential fix for the issue discussed here:

https://discuss.prosemirror.net/t/reduction-in-reconcilliation-lookahead-threshold/4184

### DIFF
--- a/src/viewdesc.js
+++ b/src/viewdesc.js
@@ -1161,7 +1161,7 @@ class ViewTreeUpdater {
         }
       }
     } else {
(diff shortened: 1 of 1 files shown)


seanchambo:
@marijnh I have changed this PR to just bump the threshold back to 5 for now. We are still investigating a way to replicate the performance issues we are having in a small script. We would like this as an interim solution though

marijnh:
Released this as 1.23.1

PR #115, opened 2021-12-23
feat: support `preserveWhitespace` in `Node.spec` to avoid dropping `newline`

lastnigtic:
https://codesandbox.io/s/prosemirror-forked-ziz0y?file=/src/index.js

As the demo shows, when typing in the `paragrph`.

All `\n` will be deleted, only `spec.code` can work.

But `code` can be used in other [situations](https://github.com/ProseMirror/prosemirror-commands/blob/master/src/commands.js#L229),  I think it would be better to support `preserveWhitespace` in `Node.spec`.

(If you can take my advice, I think I should also further update the code of `prosemirror-model`)

Is there a better solution or idea? I will be very grateful for that.

### DIFF
--- a/src/domchange.js
+++ b/src/domchange.js
@@ -39,7 +39,7 @@ function parseBetween(view, from_, to_) {
     topOpen: true,
     from: fromOffset,
     to: toOffset,
-    preserveWhitespace: $from.parent.type.spec.code ? "full" : true,
+    preserveWhitespace: $from.parent.type.spec.code ? "full" : $from.parent.type.spec.preserveWhitespace || true,
     editableContent: true,
     findPositions: find,
     ruleFromNode,


marijnh:
This seems like it would benefit from a more generic node-level setting. If you upgrade to prosemirror-model 1.16.0 and prosemirror-view 1.23.5, you should be able to include `whitespace: "pre"` in your node spec to get the effect you want. Can you let me know whether that works for you?

lastnigtic:
> This seems like it would benefit from a more generic node-level setting. If you upgrade to prosemirror-model 1.16.0 and prosemirror-view 1.23.5, you should be able to include `whitespace: "pre"` in your node spec to get the effect you want. Can you let me know whether that works for you?

I've updated the code of the upper example and it works well, thank you!

PR #116, opened 2022-01-13
Opt out of drag selection delay in Safari

luke-john-atlassian:
FIX: In Safari the drag selection delay introduced in prosemirror-view@1.19.2 (via ProseMirror/prosemirror-view@885258b) causes issues with Cell Selections (from the prosemirror-tables plugin).

The drag selection delay was introduced to resolve issues which were only present in chrome, so this change opts out of the drag selection delay for non chrome browsers.

This resolves https://github.com/ProseMirror/prosemirror/issues/1152

### DIFF
--- a/src/selection.js
+++ b/src/selection.js
@@ -42,7 +42,10 @@ export function selectionToDOM(view, force) {
 
   if (!editorOwnsSelection(view)) return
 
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks!

PR #118, opened 2022-03-10
Add empty alt attribute to img hack nodes

Billiam:
Currently, these are read as "graphic" by screen readers when present


marijnh:
Good point, thanks for the PR.

PR #119, opened 2022-04-13
Fix hidden cursor in chrome at end of line following a custom node view wrapped in a mark

Billiam:
Title is a mouthful!

Fix for https://github.com/ProseMirror/prosemirror/issues/1261

Fixes missing cursor in chrome in the above circumstance by moving the image hack node directly after the last child, rather than after the top node.

I'm not sure whether the `element.parent == this.parent` check is the correct way to decide whether to increment the current index.


marijnh:
Could you see if alternative patch https://github.com/ProseMirror/prosemirror-view/commit/a29a6b00ef20b4a937f9fe5bd968ba4b540c069e also works for you?

Billiam:
Your patch works for me, I'll close this. Thanks!

PR #120, opened 2022-04-21
Fix typo

ocavue:
`DirectEditorProps` includes `plugins` since [this commit](https://github.com/ProseMirror/prosemirror-view/commit/ffd8f3a8659ec7ae0ddeda81c31131d22d7b9536#diff-bfe9874d239014961b1ae4e89875a6155667db834a410aaaa2ebe3cf89820556R637), so it includes three fields now.


marijnh:
Indeed! Thanks for spotting that.

PR #121, opened 2022-04-25
allow editable prop to be bool

BrianHung:


marijnh:
I can see how this could be slightly more convenient, but it's easy enough to pass in `() => false`, so I don't think we really need this extension.

PR #122, opened 2022-04-26
Only flush the selection when the editor view has focus

tilgovi:
Just as 94a7758 accounts for cases where the editor blurs without losing
the selection, account for the corresponding state when flushing wherein
the editor has the selection but not the focus.

If the editor view does not have focus, then it will set the selection
when it regains focus.

### DIFF
--- a/src/domobserver.js
+++ b/src/domobserver.js
@@ -1,6 +1,6 @@
 import browser from "./browser"
 import {domIndex, isEquivalentPosition} from "./dom"
-import {hasFocusAndSelection, hasSelection, selectionToDOM} from "./selection"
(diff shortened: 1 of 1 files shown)


marijnh:
This looks reasonable!

tilgovi:
@marijnh, would it be possible to publish this as a patch version? We ran into this issue during a refactor at @nytimes, and it'd be nice for us to be able to move ahead without regressions that this change fixes for us.

marijnh:
I've tagged 1.23.3

tilgovi:
Thank you for the super quick follow-up, @marijnh!

PR #123, opened 2022-05-03
Make the `plugins` field in `DirectEditorProps` optional

ocavue:
Fix https://github.com/ProseMirror/prosemirror/issues/1265


ocavue:
This is already fixed. Closing it.

PR #124, opened 2022-05-17
Fix table duplication in clipboard

kg782:
Resolves #1270


kg782:
I found a problem of this approach so closed it.

PR #126, opened 2022-05-30
Add CSS exports to package

antimatter15:
Certain bundlers such as vite require that es module packages include all exports in the package.json, including CSS imports.


marijnh:
I've pushed attached, slightly simpler patch instead.

PR #127, opened 2022-06-04
export NodeViewConstructor and MarkViewConstructor

BrianHung:
Would be nice to expose these types for libraries that build abstractions on top of these two.


marijnh:
We'll also need doc comments on these if they are going to appear in the reference docs.

PR #129, opened 2022-06-10
Fix some types

percy507:
1. For `__endComposition`, I can't find why it can't be compiled if export its type. **Need help!!!**

2. Allow custom attrs for Decoration.widget's spec

    In [uploading image case](https://prosemirror.net/examples/upload/), we need `id` property for the widget's spec. If I want to show the upload progress, I also need add `progress` property to spec for storing the DOM which will show the progress data.

3. Fix type for `decorations`, add undefined type, because `pluginKey.getState` returns `PluginState | undefined`.

    ```ts
    new Plugin({
      ...
      props: {
        decorations(state) {
          return pluginKey.getState(state);
        },
      },
    });
    ```

### DIFF
--- a/src/decoration.ts
+++ b/src/decoration.ts
@@ -182,6 +182,8 @@ export class Decoration {
     /// Called when the widget decoration is removed as a result of
     /// mapping
     destroy?: (node: DOMNode) => void
+
+    [key: string]: any
   }): Decoration {
     return new Decoration(pos, pos, new WidgetType(toDOM, spec))
   }
--- a/src/index.ts
+++ b/src/index.ts
@@ -537,7 +537,7 @@ export interface EditorProps {
   /// `preventDefault` yourself (or not, if you want to allow the
   /// default behavior).
   handleDOMEvents?: {
-    [event in string]: (view: EditorView, event: DOMEventMap[event]) => boolean | void
+    [event in keyof DOMEventMap]?: (view: EditorView, event: DOMEventMap[event]) => boolean | void
   }
 
   /// Called when the editor receives a `keydown` event.
@@ -579,7 +579,7 @@ export interface EditorProps {
   /// Called when something is dropped on the editor. `moved` will be
   /// true if this drop moves from the current selection (which should
   /// thus be deleted).
-  handleDrop?: (view: EditorView, event: MouseEvent, slice: Slice, moved: boolean) => boolean | void
+  handleDrop?: (view: EditorView, event: DragEvent, slice: Slice, moved: boolean) => boolean | void
 
   /// Called when the view, after updating its state, tries to scroll
   /// the selection into view. A handler function may return false to
@@ -672,7 +672,7 @@ export interface EditorProps {
 
   /// A set of [document decorations](#view.Decoration) to show in the
   /// view.
-  decorations?: (state: EditorState) => DecorationSource | null
+  decorations?: (state: EditorState) => DecorationSource | null | undefined
 
   /// When this returns false, the content of the view is not directly
   /// editable.


percy507:
> 1. For `__endComposition`, I can't find why it can't be compiled if export its type. **Need help!!!**

@marijnh It seems your [buildtool](https://github.com/marijnh/buildtool/blob/main/src/buildtool.ts#L68) will stripe types which are marked as internal.  

I think it's better to export them. Because the developer may use them in some special cases. 

> eg: The `wrapInList` command use the method `compatibleContent` of NodeType which is marked as internal. And developer may use `__serializeForClipboard` or `__parseFromClipboard` for handling the clipboard.

marijnh:
The underscore prefixed ones are just used in the tests and should not be part of the doc template or public types.

What are the modification to the type of `handleDOMEvents` for?

percy507:
> The underscore prefixed ones are just used in the tests and should not be part of the doc template or public types.

OK. I will remove it later. (But in my package, I use them for toggling covert `Slice` and `dom/text`😂).

What about the the method `compatibleContent`  of NodeType? It really used in `wrapInList ` command. In my package, I don't use `prosemirror-schema-list` directly instead of writing them locally for some custom logic.

> What are the modification to the type of `handleDOMEvents` for?

```ts
handleDOMEvents?: {
  [event in string]: (view: EditorView, event: DOMEventMap[event]) => boolean | void
}
```

It seems typescript can't auto recognize the type in `DOMEventMap[event]`. On my test, all event object inside handleDOMEvents are `Event`, not the right one.

percy507:
@marijnh 

> Am I missing anything?

- Each key of `handleDOMEvents` type shoulld be optional.😂

```ts
Partial<{ [event in keyof DOMEventMap]: (view: EditorView, event: DOMEventMap[event], ) => boolean | void; }>
```

- [Allow custom attrs for Decoration.widget's spec](https://github.com/ProseMirror/prosemirror-view/pull/129/files#diff-0d652241cb103acb5218b28cbf5b5b976b10d3d2312a9e68580151de46d47c12R186)

    >     In [uploading image case](https://prosemirror.net/examples/upload/), we need `id` property for the widget's spec. If I want to show the upload progress, I also need add `progress` property to spec for storing the DOM which will show the progress data.
    
- The event type of `handleDrop` need be fixed
  - https://github.com/ProseMirror/prosemirror-view/pull/129/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80R585

marijnh:
> Each key of handleDOMEvents type shoulld be optional.

Indeed! Added a question mark there.

Could you remove the `handleDOMEvents` changes from this PR? The other changes look good.

percy507:
Ah. I just change the type of `handleDOMEvents` same to the newest one. The branch seems mergable.

percy507:
> Indeed! Added a question mark there.

@marijnh Sadly, it seems not work as expected.

image

I modify the type inside node_modules. Change the `Event` to `any` will work.

```ts
interface DOMEventMap extends HTMLElementEventMap {
    [event: string]: any;
}
```

image

percy507:
I's sorry. I can't reproduce it through online code editor. Maybe I should fix it in another way.🤯

```ts
new Plugin({
  ...
  props: {
    handleDOMEvents: {
      mouseup(view, e) {
        console.log(view, e);
      },
      keyup(view, e) {
        console.log(view, e);
      },
      dragstart(view, e) {
        console.log(view, e);
      },
    } as DirectEditorProps["handleDOMEvents"],
  },
});
```

percy507:
> image

Finally. I solve it by setting `"strictFunctionTypes": false` of tsconfig.

PR #136, opened 2022-08-09
Add failing test for decoration mapping bug

peets:
You can view this in action in a browser at https://cute-resolute-beauty.glitch.me

And the source code for the example: https://glitch.com/edit/#!/cute-resolute-beauty

Note: this doesn't fix the bug, it only provides a test case. I don't understand decoration sets well enough to try to fix this.

This is a streamlined version of the issue first reported by @DanielFelixKim in forum post https://discuss.prosemirror.net/t/decoration-mapping-through-custom-replace-step-behavior/2694

### DIFF
--- a/test/webtest-decoration.ts
+++ b/test/webtest-decoration.ts
@@ -1,7 +1,7 @@
 import ist from "ist"
-import {schema, doc, p, h1, li, ul, blockquote} from "prosemirror-test-builder"
-import {Transform, ReplaceAroundStep, liftTarget} from "prosemirror-transform"
-import {Slice, NodeRange, Node} from "prosemirror-model"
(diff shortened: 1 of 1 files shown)


marijnh:
Merged as 6ef04c322be76ace and fixed in 978011f981969c2265

ronhou:
> Merged as [6ef04c3](https://github.com/ProseMirror/prosemirror-view/commit/6ef04c322be76ace429e1124f9fc8368fb0e5792) and fixed in [978011f](https://github.com/ProseMirror/prosemirror-view/commit/978011f981969c2265030f10c023e3b05219a95c)

It should not compare oldStart & oldEnd with oldOffset, when there are more than one map.
The oldStart and oldEnd are absolute position which based on previous step maps.
But oldOffset is also an absolute position which not mapped by previous step maps.

PR #137, opened 2022-08-10
Use previous selection when focus origins from key event

jnystad:
Selection should be restored to its previous state when using tab to navigate
a page with ProseMirror and other focusable elements with selection (inputs).

This is similar to how textareas work.

Note: While it's possible to create a workaround outside of prosemirror-view by saving the selection and restoring it before/after focus, it does not seem to be a way to avoid scroll jumps when the selection is further down in the document.

The input focus handler restores the previous selection (and DOM selection) already, so ignoring focus when origin is unknown (originated outside of the editor and not via pointer click) makes it restore the previous selection correctly without causing scroll jumps.


marijnh:
This patch seems somewhat scary, because it's hard to know that it won't incorrectly ignore selections in situations where they they are actual user actions that we somehow didn't predict. Does attached patch work for you?

PR #142, opened 2022-09-21
Fix getSelection for Safari Shadow Dom 

roman-kulakov:
Use custom getSelection for Safari Shadow DOM, as native method returns incorrect anchorNode and focusNode

It fixes issues described in [1310](https://github.com/ProseMirror/prosemirror/issues/1310)

### DIFF
--- a/src/dom.ts
+++ b/src/dom.ts
@@ -94,3 +94,9 @@ export function keyEvent(keyCode: number, key: string) {
   ;(event as any).key = (event as any).code = key
   return event
 }
+
+export function deepActiveElement(doc: Document) {
+  let elt = doc.activeElement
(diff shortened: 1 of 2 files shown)


marijnh:
This is a great start, but I don't want to include the hack where you create an object with the methods and properties of the native selection copied into it. I'd prefer to do what CodeMirror does and distinguish between selection range objects (just `focusNode`, `focusOffset`, `anchorNode`, `anchorOffset`) and selection objects that are going to have their methods called throughout the code base. Most places that need this just need the range, for which we'd add a custom getter function, and the few that do need to update the selection will have to work with two objects, the range + the native selection object.

marijnh:
I've gone ahead and implemented what I meant in 64a96e51be902fdcd96d210cf0241293750b16f1

PR #143, opened 2022-10-04
chrome => isChrome

eric-burel:
Fixes ProseMirror/prosemirror/issues/1318

Variables defined in "browser" will end up in browser with the exact same name, which is problematic for "chrome". Renaming to isChrome should fix the issue.


marijnh:
Why is this problematic?

PR #144, opened 2022-10-08
fix(clipboard-readHTML-windows): Fix readHTML issue on windows

hajikanooo:
Hi marijnh,
On mac,the copied data looks like:
`<meta charset='utf-8'><table data-pm-slice="1 1 -2 []"><tbody><tr><td class="PROSEMIRROR_TABLE_TD_CLS" align="left"><p>123</p></td></tr><tr><td class="PROSEMIRROR_TABLE_TD_CLS" align="left"><p>123</p></td></tr></tbody></table>`
However,on windows,the data looks like:
`<html><body><!--StartFragment--><table data-pm-slice="1 1 -2 []"><tbody><tr><td class="PROSEMIRROR_TABLE_TD_CLS" align="left"><p>123</p></td></tr><tr><td class="PROSEMIRROR_TABLE_TD_CLS" align="left"><p>123</p></td></tr></tbody></table><!--EndFragment--></body></html>`
This is the problem and the solution is included in this pr.
Thanks !

### DIFF
--- a/src/clipboard.ts
+++ b/src/clipboard.ts
@@ -202,6 +202,8 @@ function detachedDoc() {
 function readHTML(html: string) {
   let metas = /^(\s*<meta [^>]*>)*/.exec(html)
   if (metas) html = html.slice(metas[0].length)
+  const fragments = html.match(/<!--StartFragment-->(.*)<!--EndFragment-->/); // for windows
+  if (fragments) html = fragments[1]
   let elt = detachedDoc().createElement("div")
   let firstTag = /<([a-z][^>\s]+)/i.exec(html), wrap
   if (wrap = firstTag && wrapMap[firstTag[1].toLowerCase()])


hajikanooo:
We have solved this problem by transformPastedHTML.

marijnh:
What issue does the added HTML cause for you? It looks like it would just be ignored by the DOM parser.

hajikanooo:
> What issue does the added HTML cause for you? It looks like it would just be ignored by the DOM parser.

Thank you for replying.The exact issue is the pasted action from some table cells.If you copy the table cells,and paste them in the editor,that will be worked on mac,but not windows.This is the problem.
On the other way,in this function `readHTML`,you add the `metas` reg to delete the meta tags,which will just be the mac.But on windows,the metas is not exist,but the added HTML.
So my commit code about `fragments` reg,is just do the same thing like the metas reg.
The DOM parser can fix this problem too,but why don't we do this thing in `readHTML` funtion?Just the same thing like `metas` reg!

Have a nice day!

marijnh:
I could reproduce the issue, and I've pushed an alternative patch that solves it in a more general way.

hajikanooo:
> I could reproduce the issue, and I've pushed an alternative patch that solves it in a more general way.

 Thank you most sincerely

PR #145, opened 2022-10-20
FIX: Keep multiple line breaks when pasting text

xiaweiss:
when paste text from bear editor: https://bear.app/

image

content: `Text\n\nText`

browser contenteditable show:
```
Text

Text
```

image

but prosemirror editor show:
```
Text
Text
```

image

so need to keep multiple line breaks


marijnh:
This works this way intentionally, and is not going to change. You can set up a custom [`clipboardTextParser`](https://prosemirror.net/docs/ref/#view.EditorProps.clipboardTextParser) to get different behavior in your editor.

xiaweiss:
Ok. Thank you.

PR #147, opened 2022-11-04
DecorationGroup.from accepts DecorationSet | DecorationGroup

jonathonherbert:
An attempt at fixing https://github.com/ProseMirror/prosemirror/issues/1329.

For the test, I couldn't think of another way to produce DecorationGroups, as every interface that creates them is internal – be interested to know if there's a better approach.


marijnh:
I've pushed c3d3ccdc3477360c11e7d00254ffe33537a3e8c7, which does the same thing in a different way, instead. Merged test case.

PR #148, opened 2022-11-25
Support auto scroll while dragging on safari and improve the scroll interaction on other browsers

percy507:


percy507:
https://github.com/ProseMirror/prosemirror/issues/1339

marijnh:
This will leak event handlers and interfere with all drag events on the page, it seems.

percy507:
True. I make an another commit to solve this. https://github.com/ProseMirror/prosemirror-view/pull/148/commits/718808471c892bf1d51aa28609db93f6f469f4e5

marijnh:
Sorry, this still feels too messy to merge. There might be other drop targets next to the editor, in which case scrolling the editor when dragging something near it seems inappropriate. This could be implemented as a plugin, without modifying the editor core. And since most browsers do this natively already, I don't think it's worth adding extra, potentially problematic code to the library for.

percy507:
It's ok. The code of this PR may inspire someone who face same problem.

PR #149, opened 2023-01-05
Fix composition after nodeview with mark(inclusive is false) will get…

cuppi:
Current behavior: In the nodes on the right side with mark (mark inclusive attribute is false) perform the IME input will cause an exception is thrown, and the input character will disappear。

Expected behavior: The words entered by the IME procedure can be displayed to the right of the node without throwing an exception。

Demo to reproduce the issue: [Glitch](https://glitch.com/edit/#!/spotty-octagonal-salt).
Screen Recording: .
Error Snapshot: 

Reason: When renderDescs renders MarkViewDesc's parent NodeViewDesc, MarkViewDesc's dom and CompositionViewDesc's dom point to the same node, The composition node was accidentally deleted during the subsequent MarkViewDesc render.

Solution: During Composition, the MarkViewDesc node around CompositionViewDesc is recreated to ensure that the DOM and contentDOM point to different nodes from the CompositionViewDesc

### DIFF
--- a/src/viewdesc.ts
+++ b/src/viewdesc.ts
@@ -791,6 +791,17 @@ export class NodeViewDesc extends ViewDesc {
 
     // Patch up this.children to contain the composition view
     this.children = replaceNodes(this.children, pos, pos + text.length, view, desc)
+    // Force re-create markdesc dom to prevent renderDesc remove composition view
+    const changeDescIdx = new Set<number>();
+    this.children.forEach((desc, idx) => {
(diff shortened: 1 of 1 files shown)


cuppi:
> Could you see if patch [67a87c2](https://github.com/ProseMirror/prosemirror-view/commit/67a87c2e63fdc085233162df7e9eada643afd070) also works for you?

Thank you for your reply, this patch also works.

marijnh:
Great to hear.

PR #150, opened 2023-01-09
Can't input with composing when the first node in selection range isn't editable

lastnigtic:
If the attribute `contenteditable` of the first content in selection range is `false`, it can't input with composing.
image

It can be reproduced at [examples/schema](https://prosemirror.net/examples/schema/)

https://user-images.githubusercontent.com/17680884/211236509-584b9a4d-4e94-4083-b828-c1f2666b70fa.mov

The behavior is the same when the first content in selection range is `block` [code example](https://codesandbox.io/s/prosemirror-basic-example-with-parcel-forked-bttq0q?file=/src/index.js).

Try to solve this problem, if there is a better solution, please share your ideas, thanks!

### DIFF
--- a/src/input.ts
+++ b/src/input.ts
@@ -443,6 +443,17 @@ function inOrNearComposition(view: EditorView, event: Event) {
   return false
 }
 
+function isHeadContentAtom(selection: Selection) {
+  let firstChild = selection.$from.nodeAfter;
+  while (firstChild) {
+    if (firstChild.isAtom && !firstChild.isText) {
+      return true;
+    }
+    firstChild = firstChild.firstChild
+  }
+  return false;
+}
+
 // Drop active composition after 5 seconds of inactivity on Android
 const timeoutComposition = browser.android ? 5000 : -1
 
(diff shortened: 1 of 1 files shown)


marijnh:
Please tell us which browser and which IME this is happening with. I can't reproduce the problem with IBus on Linux Firefox or Chrome.

lastnigtic:
> Please tell us which browser and which IME this is happening with. I can't reproduce the problem with IBus on Linux Firefox or Chrome.

I'm using `MacOs Montery 12.5.1`.

For me it's reproducible in  `Chrome 108.0.5359.124` and `Firefox 108.0.2`, but works fine in `Safari`.

marijnh:
I can reproduce this on MacOS, but it seems to be more of an OS (or browser) issue than something ProseMirror can help. The same issue occurs if you make a plain contenteditable element and put the selection across an uneditable element inside of it. The IME state seems to be screwed up until you blur and refocus the editable element, in fact—even when putting the selection away from the uneditable elements, you can't compose normally.

lastnigtic:
Got it, so your suggestion is that I make this change a special treatment for my own use, or is there another better way to deal with it?

marijnh:
I don't really have a suggestion for how to fix this. If you find a workaround, I'd be interested in that, but this looks like it's happening in OS-level components that JavaScript code can't really control. You could also report bugs to Chrome and Firefox, of course, but that's not likely to yield short-term results.

PR #151, opened 2023-02-15
Prevent model selection synchronizing to DOMSelection during selection.

cuppi:
Current behavior: Now there are some selection issues that are bothering me, e.g.
A NodeView with margin(margin-left and margin-right), when select from NodeView's right to left,  the selection is collapsed accidentally.

A NodeView that regardless of margin, when select from NodeView's right to left, and shuffling between text and NodeView, the anchor of the selection was be collapse to NodeView left. 

A NodeView that has content in the schema but does not set with contentDOM, when select from NodeView's right to left, and shuffling between text and NodeView, the anchor of the selection was be collapse to NodeView left. 

Expected behavior: The selection should not be affected by the element style and "anchor" should be fixed while moving the "focus"

Through some experiments, I found that when I set the selection in the selectionChange callback triggered by mouse interaction, when the set selection is not the "legal" position considered by the browser, the browser may readjust the selection. Therefore, I boldly submitted a PR  -  In the callback of selectionChange triggered by mouse interaction, do not synchronize selection from model to DOM. I think there might be some unexpected effects, so I'm looking forward to your insight.

Demo to reproduce the issue: [Glitch](https://glitch.com/edit/#!/curse-organized-silicon).

### DIFF
--- a/src/domobserver.ts
+++ b/src/domobserver.ts
@@ -43,6 +43,7 @@ export class DOMObserver {
   currentSelection = new SelectionState
   onCharData: ((e: Event) => void) | null = null
   suppressingSelectionUpdates = false
+  lockDOMSelectionSync = false
 
   constructor(
     readonly view: EditorView,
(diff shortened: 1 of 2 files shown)


marijnh:
This is unlikely to be merged as-is, since indeed, it'll probably break some stuff. Could you set up a demo that allows me to reproduce the behavior in your screencasts? I assume this is Chrome, behaving like that?

PR #152, opened 2023-02-17
bugfix: avoid bug if domSelection is not displayed in the webpage

NomadeLampe:
Hello, during my work I use [Kendo Editor](https://www.npmjs.com/package/@progress/kendo-angular-editor) which use this component. It appear that from time to time on Chrome we have the following error `TypeError: Cannot read properties of null (reading 'focusNode')\n at EditorView.updateStateInner`.

I believe that this small change would prevent this bug.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -195,7 +195,7 @@ export class EditorView {
         // Chrome sometimes starts misreporting the selection, so this
         // tracks that and forces a selection reset when our update
         // did write to the node.
-        let chromeKludge = browser.chrome ? (this.trackWrites = this.domSelectionRange().focusNode) : null
+        let chromeKludge = browser.chrome ? (this.trackWrites = this.domSelectionRange()?.focusNode) : null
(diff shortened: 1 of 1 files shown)


marijnh:
According to the [spec](https://w3c.github.io/selection-api/#dom-document-getselection) `Document.getSelection` (which is where this value originates) should always return a `Selection` object unless the document has no browsing context (isn't any window's actual active document). If that is a thing that actually happens in the wild, we'll need to adjust more than just this to account for it. Do you have a way to reproduce this crash or more information about what might be going on when it happens?

NomadeLampe:
I have notice this typing from [typescript](https://github.com/microsoft/TypeScript/blob/main/lib/lib.dom.d.ts#L4769) where it return `Selection | null` but in [index.ts](https://github.com/ProseMirror/prosemirror-view/blob/master/src/index.ts#L473) the typing looks wrong.

Regarding the way to reproduce it, I will try but I can't copy any code. It happen when use multiple instance of kendo-editor and try to interact with the first one while having a pop-in hiding the editor.

PR #153, opened 2023-02-21
Clarify getPos typings in NodeView constructor

sarahriouslydt:
Hello!
Submitting a small PR to clarify the typings of the getPos() function provided in constructors of `NodeView`s.
The typing states that it only returns a well-defined number, but in-fact [the implementation ](https://github.com/ProseMirror/prosemirror-view/blob/master/src/viewdesc.ts#L640) can fall through and return an `undefined` position.

We ran into a case with this where a callback in React was attempting to set the document selection using the `getPos` of a Node view after the node had been removed from the document. This resulted in trying to set a selection using an undefined pos which causes a `Position out of Range` error to be thrown.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -546,7 +546,7 @@ function checkStateComponent(plugin: Plugin) {
 
 /// The type of function [provided](#view.ViewProps.nodeViews) to
 /// create [node views](#view.NodeView).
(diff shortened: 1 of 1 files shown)


marijnh:
That seems reasonable, though it may cause type errors in existing project.

PR #154, opened 2023-02-27
Update broken link (ViewProps -> EditorProps)

shreevatsa:
Found a few broken links while reading the documentation at https://prosemirror.net/docs/ref/


marijnh:
Thanks for spotting those! The docs will be regenerated on the next prosemirror-view release.

PR #155, opened 2023-03-15
FIX: Fix an issue with cursor jump in table cell

deniszografski:
@marijnh 
I encountered a problem when the user wants to choose a different cell and in condition that the previous one has some content. 

I also recorded a short gif demonstrating the problem 
ezgif com-optimize
 
and I also made a solution as part of this PR.  

P.S.
It may not be in line with the best practices of the whole lib but it can serve as an inspiration for a potential fix. Thanks in advance.

Stack: 

- browser: Google Chrome **Version 111.0.5563.64** 
- prosemirror-view- **1.30.2** 
- prosemirror-collab - **1.30.0** 
- prosemirror-tables - **1.2.5**

### DIFF
--- a/src/input.ts
+++ b/src/input.ts
@@ -388,7 +388,7 @@ class MouseDown {
                 // (hidden) cursor is doesn't change the selection, and
                 // thus doesn't get a reaction from ProseMirror. This
                 // works around that.
-                (browser.chrome && !this.view.state.selection.visible &&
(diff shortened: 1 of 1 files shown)


marijnh:
Could you describe the problem this is intended to fix in a bit more detail?

marijnh:
I cannot reproduce the problem you are describing. This happens in Chrome, right, when you click around to move the cursor between adjacent empty cells?

PR #158, opened 2023-05-08
Allow translation if contenteditable is set to false

seeARMS:
Translate breaks prosemirror when someone begins editing (https://github.com/ProseMirror/prosemirror/issues/1074 was fixed by setting `translate=no` on the view).

We would like the already-rendered and uneditable content to be translatable, though (https://github.com/ueberdosis/tiptap/issues/3826).

This PR fixes https://github.com/ueberdosis/tiptap/issues/3826 by setting the `translate` property based on the `editable` property of the view.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -478,7 +478,7 @@ function computeDocDeco(view: EditorView) {
   let attrs = Object.create(null)
   attrs.class = "ProseMirror"
   attrs.contenteditable = String(view.editable)
-  attrs.translate = "no"
+  attrs.translate = view.editable ? "no" : "yes"
 
   view.someProp("attributes", value => {
     if (typeof value == "function") value = value(view.state)


marijnh:
You should be able to use the [`attributes`](https://prosemirror.net/docs/ref/#view.EditorProps.attributes) prop to override the value of the `translate` attribute. I don't think "if not editable, has to be translateable" always holds (being editable might be dynamic, for example), and hard-wiring it like that will probably surprise more people than it helps. So I recommend using the `attributes` prop approach.

seeARMS:
Setting `attributes` doesn't work currently since this line checks if the attribute value is already defined on `attrs`, which it is in the case of `translate`: https://github.com/seeARMS/prosemirror-view/blob/a2d35e2da697418ab6ac30dd9765557267113d28/src/index.ts#L491

I can modify that line to allow `translate` to override?

marijnh:
Oh, that's a good point. That isn't intentional behavior. Attached patch modifies the logic to allow `translate` to be overridden.

PR #159, opened 2023-05-28
mark `MarkView` as a single interface

bvanjoi:
Make the return type of `MarkViewConstructor` more concise


marijnh:
This doesn't seem a big enough win to justify exporting another type.

PR #161, opened 2023-11-02
Incorrect Step for some unicode characters during composition

saranrapjs:
My coworkers have built an integration testing rig (using Puppeteer) to simulate a lot of browser clients collaborating on a single ProseMirror document, using `prosemirror-collab`. One bug we've caught by way of this testing rig seems like it might originate within ProseMirror code itself, and specifically within `prosemirror-view`. Here's a description of the approximate sequence of events:

|client | typed emoji input|unicode equivalent for emoji|resulting step for input|doc on screen|
|---|---|---|---|---|
| A | 👽 | `\ud83d\udc7d` | 👽 | 👽 |
| B | 💎 | `\ud83d\udc8e` | `\udc8e\ud83d` | 💎👽 |

Where the thing to note is that the doc on screen (e.g. the `EditorView.state.doc`) contains a correct/expected emoji sequence of "gemstone + alien", but the step produced from inserting the gemstone emoji includes a garbled unicode sequence that's effectively the equivalent of if you'd executed the expression `'💎👽'.slice(1, 3)` to produce its slice content.

This happens relatively rarely with our testing rig, but not never. The attached branch reproduces the issue with steps data within the existing `prosemirror-view` webtests:

Screenshot 2023-11-02 at 12 56 52 PM

I'd be curious whether this is a known issue with composition input handling, and/or whether you can recommend a particular fix. (We'd be happy to help execute a fix, too, but I wanted to start with sharing a reproduction of the bug first to make sure we're on the right track for exercising the bug itself!)

### DIFF
--- a/test/webtest-composition.ts
+++ b/test/webtest-composition.ts
@@ -3,6 +3,7 @@ import ist from "ist"
 import {Decoration, DecorationSet, __endComposition, EditorView} from "prosemirror-view"
 import {EditorState, Plugin} from "prosemirror-state"
 import {tempEditor, requireFocus, findTextNode} from "./view"
+import { ReplaceStep } from "prosemirror-transform"
 
 function event(pm: EditorView, type: string) {
   pm.dom.dispatchEvent(new CompositionEvent(type))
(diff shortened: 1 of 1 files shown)


marijnh:
It seems that the diffing of the input, comparing the following sequences, finds a valid diff in the middle of two surrogate pairs, because both these Emoji start with in the same low half.

Before: d83d dc7d
After: d83d dc8e d83d dc7d

Both 'insert "d83d dc8e" at start' and 'insert "dc8e d83d" in the middle' form valid patches here, on the JavaScript string level. But, yeah, the latter is pretty dodgy, and I guess it caused some kind of problem in your process, that you're reporting it?

I could add some logic in the DOM change parser that biases diffing towards diffs that don't start in the middle of a surrogate. Does that sound like it'd help for you?

saranrapjs:
> Both 'insert "d83d dc8e" at start' and 'insert "dc8e d83d" in the middle' form valid patches here, on the JavaScript string level. But, yeah, the latter is pretty dodgy, and I guess it caused some kind of problem in your process, that you're reporting it?

That's right — while it's probably rare that people in real world usage are pasting in emojis which happen to have overlapping unicode subsequences, it's still possible (we certainly do have instances where people type in a number of different emoji in a row).

> I could add some logic in the DOM change parser that biases diffing towards diffs that don't start in the middle of a surrogate. Does that sound like it'd help for you?

I think this makes sense to me!

marijnh:
Attached patch should help. Changes were already aligned to the selection when possible, so in actual user-driven situations, where the cursor was next to the change this would probably never happen. But with this change, there's an extra check.

saranrapjs:
I'm now understanding why you'd asked about our process (and maybe implicitly, whether it's user-driven). I should have mentioned before that I think this happens with simulated browsers because puppeteer's "input" API tries to be smart and will dispatch different browser events over Chrome's devtools protocol [depending on whether or not it sees the input as keyboard-y](https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/src/cdp/Input.ts#L191-L200). So all non-ascii character inputs get the IME treatment (I suspect without a selection change), even if they wouldn't trigger composition input events in real world use.

PR #163, opened 2023-12-01
Create unit tests for a decoration mapping bug

becha2:
Hello,
Our team has come across a bug with the DecorationSet map function. I've created a test suite where some of the tests do fail. I wasn't sure how to best integrate this with the existing tests, but I at least wanted to bring this to your attention in hopes that the bug could be fixed.

Our scenario is we are trying to insert multiple nodes throughout the doc in a single transaction. Then we undo the transaction. On undo, we notice the resulting decoration positions are off.  

The tests I created mimick this scenario by doing the following:
1. Start with the original doc `beforeDoc`
2. Apply a mapping `actionMapping`, representing our transaction for inserting nodes
3. Apply a mapping `undoMapping`, representing our transaction for undoing the previous action

Of the four test cases I created, only "test case 2" passes all the tests. What surprises me is that "test case 1" and "test case 2" have the same decoration positions but the difference is the placements of the inserted nodes in between decorations.
pm-deco-test

### DIFF
--- a/test/test-decoration-mapping.ts
+++ b/test/test-decoration-mapping.ts
@@ -1,10 +1,10 @@
 import assert from 'assert';
 import { Decoration, DecorationSet } from 'prosemirror-view';
-import { StepMap, Mapping } from 'prosemirror-transform';
+import { Transform } from 'prosemirror-transform';
 import { Node as PmNode } from 'prosemirror-model';
 import { doc, p, blockquote } from 'prosemirror-test-builder';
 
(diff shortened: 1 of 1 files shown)


marijnh:
Why are you creating the mappings manually, rather than deriving them from steps? That makes it rather hard to verify they are correct.

becha2:
> Why are you creating the mappings manually, rather than deriving them from steps? That makes it rather hard to verify they are correct.

ah, i created the mapping objects based on what I had observed while I was debugging.
I updated the PR to use a Transform instead. Would this be better?

PR #164, opened 2023-12-27
fixes memory leak related to reused range

gwidonw:
Hi team 👋 , 
I create PM instance in a dialogue, so in my use case PM instance is created and destroyed quite often. I've noticed that after destroying PM, some references to DOM are still kept in memory - this is due to `reusedRange` not being cleared. 

This small change below should help to remove the memory leak caused by the above.

### DIFF
--- a/src/dom.ts
+++ b/src/dom.ts
@@ -29,6 +29,10 @@ export const textRange = function(node: Text, from?: number, to?: number) {
   return range
 }
 
+export const clearReusedRange = function() {
+  reusedRange = null;
+}
+
 // Scans forward and backward through DOM positions equivalent to the
 // given one to see if the two are in the same place (i.e. after a
 // text node vs at the end of that text node)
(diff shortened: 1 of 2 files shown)


marijnh:
>  some references to DOM are still kept in memory

The code looks like it will hold onto a single DOM text node (and never more, since the next editor will continue to reset the existing range). Is this really causing any concrete problem for you?

gwidonw:
Yes, because of this single node, the entire dialogue DOM structure is not garbage collected. In my case, the dialogue with the prose mirror editor is quite complex, and it ends up keeping a lot of DOM detached elements and event listeners in memory. Cleaning up the `reusedRange` solves this problem. 

Screenshot 2023-12-27 at 12 29 20

gwidonw:
@marijnh, just wanted to bump this PR. Can we merge it or can I help anyhow to make it happen? Thanks

marijnh:
I guess you're ending up with the range pointing at a text node that has a `pmViewDesc` property, which can indeed keep an entire editor alive. Merged and released as 1.33.2.

PR #165, opened 2024-02-14
expose `forChild` of `DecorationSource`

twrichards:
In https://github.com/guardian/prosemirror-elements we have the concept of a 'repeater' which when we repeat inner prosemirror instances passing outer decorations (see https://discuss.prosemirror.net/t/is-it-possible-to-get-hold-of-decorations-that-cover-the-range-managed-by-a-nodeview-in-that-nodeview/3544) we get various catastrophic RangeErrors.

https://github.com/guardian/prosemirror-elements/pull/349 effectively fixes by extracting the decorations relevant to the repeater to pass down to the views for them to treat as 'outer' decorations, but it does so be using the currently internal `forChild` of `DecorationSource`.

### DIFF
--- a/src/decoration.ts
+++ b/src/decoration.ts
@@ -258,7 +258,7 @@ export interface DecorationSource {
   map: (mapping: Mapping, node: Node) => DecorationSource
   /// @internal
   locals(node: Node): readonly Decoration[]
-  /// @internal
+  /// Extract a DecorationSource containing decorations for the given child node at the given offset.
   forChild(offset: number, child: Node): DecorationSource
   /// @internal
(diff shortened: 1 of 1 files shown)


marijnh:
This looks unproblematic. Could you use "child" node rather than just "node" in the doc comment to make it clear it only works on direct children? Also "Exract"→"Extract".

twrichards:
> This looks unproblematic. Could you use "child node" rather than just "node" in the doc comment to make it clear it only works on direct children? Also "Exract"→"Extract".

Great, thank you @marijnh . I've made the changes you suggested.

PR #166, opened 2024-02-15
expose `forChild` of `DecorationSet`

twrichards:
Small corrective follow-up to https://github.com/ProseMirror/prosemirror-view/pull/165 since `forChild` was still marked as `internal` in `DecorationSet` (one of the implementations of `DecorationSource`) which I missed, and results in things like...
image

### DIFF
--- a/src/decoration.ts
+++ b/src/decoration.ts
@@ -413,7 +413,7 @@ export class DecorationSet implements DecorationSource {
     return local.length || children.length ? new DecorationSet(local, children) : empty
   }
 
-  /// @internal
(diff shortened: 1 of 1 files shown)


twrichards:
Sorry @marijnh - I hadn't realised one of the implementations of `DecorationSource` also had `forChild` marked as `internal`.

AFAIK there's [72 hours where you could unpublish](https://docs.npmjs.com/unpublishing-packages-from-the-registry) 1.33.0 to avoid this causing any problems for people.

marijnh:
Tagged 1.33.1. No need to unpublish. People installing or upgrading now will get the fixed version.

PR #168, opened 2024-04-27
use `transformCopied` in `dragstart` handler

markcrispo:
### Description of issue
When you drag and drop content, `transformCopied` and `transformPasted` are both called, but upon dropping content `transformPasted` receives the original slice instead of the transformed slice.

`editHandlers.dragstart` currently calls `transformCopied` via the `serializeForClipboard` function, but the transformed slice isn't attached to `view.dragging`. So when `editHandlers.drop` calls `transformPasted`, it receives the original (unmodified) slice.

I would have expected `transformPasted` to receive the transformed slice returned from `transformCopied`, which is how the editor behaves with cut and paste. 

### Description of fix
Inside the `dragstart` handler, my PR grabs the transformed slice from `serializeForClipboard` so it can be attached to `view.dragging = new Dragging(...)`

### DIFF
--- a/src/clipboard.ts
+++ b/src/clipboard.ts
@@ -36,7 +36,7 @@ export function serializeForClipboard(view: EditorView, slice: Slice) {
   let text = view.someProp("clipboardTextSerializer", f => f(slice, view)) ||
       slice.content.textBetween(0, slice.content.size, "\n\n")
 
-  return {dom: wrap, text}
+  return {dom: wrap, text, slice}
 }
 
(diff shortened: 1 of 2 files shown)


marijnh:
Thanks, that indeed makes more sense.

markcrispo:
Thanks for reviewing @marijnh! I'd like to pull in this fix since it'd let me get rid of a hacky workaround. When are new versions typically released?

marijnh:
Sure. I've tagged 1.33.6

PR #169, opened 2024-07-18
Pass attrs to renderSpec on creation of nodes and marks

sjohnson18:
Enables XSS of attrs to be checked on initial document load. Relies on this [PR](https://github.com/ProseMirror/prosemirror-model/pull/84) to go in first.


marijnh:
Oh, right, this is pretty essential to that whole protection. I've pushed an alternative patch that does some dodgy casting in order to not rely on the public interface of `renderSpec` exposing this kludge in 821031f94a5ad331c15b98cdf0fc46bfc62818d2

PR #170, opened 2024-08-01
Fix: add sel checks

xoxoxo:
For some reasons sel can be undfiened and this can cause a crash

Screenshot 2024-08-01 at 11 37 49

### DIFF
--- a/src/domobserver.ts
+++ b/src/domobserver.ts
@@ -141,7 +141,7 @@ export class DOMObserver {
   }
 
   ignoreSelectionChange(sel: DOMSelectionRange) {
-    if (!sel.focusNode) return true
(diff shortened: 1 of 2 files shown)


marijnh:
Do you know in what situation, and on which browser, you are getting null here? According to the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/API/Window/getSelection) this should only happen in cases like a disconnected iframe or a document that's not associated with a browser window.

xoxoxo:
Ah yes makes sense I actually use tiptap as editor inside iframe as its needed to isolate styles etc.

marijnh:
I didn't manage to reproduce this situation, but I guess it's not too difficult to handle. Done in attached patch.

PR #171, opened 2024-08-21
Export `DecorationGroup` so that it can be used by consumers

rhystmills:
Export `DecorationGroup`, so that consuming packages can make use of its type.

### Why?
We want `DecorationGroup` to be exportd so that we can make a plugin aware that it may receive either a `DecorationSet` or a `DecorationGroup`. `DecorationGroup` can't be used as a type currently because it isn't exported from prosemirror-view.

It seemed more useful to export the class than to make an interface mirroring the class, because the interface and class would then need to be updated in tandem.

In our case, we are handing decorations from a parent editor to a nested editor controlled by [our plugin](https://github.com/guardian/prosemirror-elements/pulls). We receive decorations from the parent editor, and the plugin receives either a `DecorationGroup` or a `DecorationSet` ([see code here](https://github.com/guardian/prosemirror-elements/blob/0deb32f485e5e04768eaa6e20c8711794d160fcb/src/plugin/fieldViews/ProseMirrorFieldView.ts#L354)).  In order to find only the relevant decorations to the nested field, we access the `find` method in the case of`DecorationSet`, or access the underlyng `members` in a `DecorationGroup` and find the relevant decorations for those members.

We currently maintain [our own type declaration](https://github.com/guardian/prosemirror-elements/blob/main/src/declarations/prosemirror-view.d.ts) for `DecorationGroup` to make it possible to use in TypeScript, but we'd rather use the interface from prosemirror-view directly.

### DIFF
--- a/src/decoration.ts
+++ b/src/decoration.ts
@@ -482,7 +482,7 @@ const empty = DecorationSet.empty
 // An abstraction that allows the code dealing with decorations to
 // treat multiple DecorationSet objects as if it were a single object
 // with (a subset of) the same interface.
-class DecorationGroup implements DecorationSource {
+export class DecorationGroup implements DecorationSource {
   constructor(readonly members: readonly DecorationSet[]) {}
 
   map(mapping: Mapping, doc: Node) {
--- a/src/index.ts
+++ b/src/index.ts
@@ -13,7 +13,7 @@ import {readDOMChange} from "./domchange"
 import {DOMSelection, DOMNode, DOMSelectionRange, deepActiveElement, clearReusedRange} from "./dom"
 import * as browser from "./browser"
 
-export {Decoration, DecorationSet, DecorationAttrs, DecorationSource} from "./decoration"
+export {Decoration, DecorationSet, DecorationAttrs, DecorationSource, DecorationGroup} from "./decoration"
 export {NodeView} from "./viewdesc"
 
 // Exported for testing


marijnh:
This isn't really supposed to be public. How is your plugin receiving this value?

rhystmills:
We are handling nested editors in our plugin, where we manage quite a lot of things in the `NodeView` for each node, with each node representing a text field. The decorations are the `innerDecorations` value passed to the `NodeView`s in the plugin via the [`update`](https://github.com/ProseMirror/prosemirror-view/blob/689e7d8034408694dcca05782e53b9d3fcc20c54/src/viewdesc.ts#L46) method. These are typed as a `DecorationSource` - in practice we have found that sometimes this is a `DecorationSet`, and sometimes it's a `DecorationGroup`. I'm not completely sure when each case occurs, but I have a feeling we get a `DecorationGroup` more often when the node contains both inline and widget decorations.

marijnh:
Would adding a method to `DecorationGroup` that allows you to enumerate the `DecorationSet` values it contains solve this for you?

marijnh:
Does attached patch look good?

rhystmills:
Yes, looks good to me, thanks.

marijnh:
Tagged in release 1.34.0

rhystmills:
Closed because @marijnh released an alternative fix to our problem.

PR #172, opened 2024-09-10
Clear trackWrites on destroy to prevent memory leak

schontz:
I did some memory profiling and came across an orphaned editor in a detached DOM node. It looks to be related to trackWrites:

Screenshot 2024-09-10 at 12 08 13 PM

This patch clears trackWrites on destroy to prevent this case.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -458,6 +458,7 @@ export class EditorView {
     }
     this.docView.destroy()
     ;(this as any).docView = null
+    this.trackWrites = null
     clearReusedRange();
   }
 


marijnh:
This is a link from the editor object, which you supposedly dropped references to, to a DOM node. I do not see how this would be able to introduce a leak. Do you have an example script that demonstrates this?

schontz:
I'll see if I can make a minimal repro of it. I was debugging our app in production, which uses tiptap on top of prose mirror.

schontz:
The root cause of the leak I was experiencing was due to tiptap code: https://github.com/ueberdosis/tiptap/issues/5654

But there is still an interesting condition that we may want to address here that actually is not related to destroying. This code appears to run before the actual editor is created, such that it grabs DOM nodes that may get replaced by the editor:

https://github.com/ProseMirror/prosemirror-view/blob/929127026fd8e2a18c1fcb97cd75aa7f0f4d2be1/src/index.ts#L199

In the linked ticket, I added a timeout to check on the kludge on the next tick:

```javascript
setTimeout(() => console.log('Is trackWrites still in the doc?', this.trackWrites?.getRootNode()?.isConnected));
```
It was logging false whenever a new editor was replacing an old editor at the same location in the DOM and the cursor was in the old editor.

I switched `trackWrites` to use WeakRef and it fixed the issue of referencing an orphaned DOM node.

```javascript
trackWrites = null;
get trackWrites() { return this.trackWritesRef?.deref(); }
set trackWrites(value) {
  if (value) this.trackWritesRef = new WeakRef(value);
  else this.trackWritesRef = null;
}
```

WeakRef is [well-supported](https://caniuse.com/mdn-javascript_builtins_weakref), but clearing `trackWrites` if it is disconnected via the immediate timeout should also suffice. And clearing it on destroy is also a good practice, so we are not hanging onto DOM nodes any longer than we have to.

schontz:
You can see my foray into prosemirror-view in this patch:

https://github.com/schontz/tiptap-react-leak/blob/c132b903a9069dd3bb8782f180fbbe3eada905dd/patches/prosemirror-view%2B1.34.3.patch

PR #174, opened 2024-11-01
Add MarkView.destroy

ocavue:
This PR implements the `destroy()` method for mark views, which is required for rendering mark with other UI framework like React. See also https://discuss.prosemirror.net/t/support-destroy-method-in-markview/7855

### DIFF
--- a/src/viewdesc.ts
+++ b/src/viewdesc.ts
@@ -589,7 +589,7 @@ class CompositionViewDesc extends ViewDesc {
 // some cases they will be split more often than would appear
 // necessary.
 class MarkViewDesc extends ViewDesc {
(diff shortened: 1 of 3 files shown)


src/viewdesc.ts
```diff
@@ -971,7 +993,7 @@ class CustomNodeViewDesc extends NodeViewDesc {
   }
 
   destroy() {
-    if (this.spec.destroy) this.spec.destroy()
+    this.spec.destroy?.()
```
marijnh, line 996: TypeScript, with our build settings, still produces much worse code for a `?.()` call than this conditional. Could you rever this and also use an `if` in `MarkView.destroy`?

ocavue: Sure. I have reverted in the latest commit.

marijnh:
Thanks!

PR #175, opened 2024-11-13
Fixed error, cannot read properties of undefined reading .nodeType

nanov94:
### What?
Get error: `TypeError: Cannot read properties of undefined (reading 'nodeType')`

### Why?
In some cases `node` does not have children and as a result we get property `nodeType` from `undefined`

### How?
Checked existing of `prev`

### DIFF
--- a/src/domcoords.ts
+++ b/src/domcoords.ts
@@ -300,11 +300,14 @@ export function posAtCoords(view: EditorView, coords: {top: number, left: number
           offset++
       }
     }
-    let prev
     // When clicking above the right side of an uneditable node, Chrome will report a cursor position after that node.
-    if (browser.webkit && offset && node.nodeType == 1 && (prev = node.childNodes[offset - 1]).nodeType == 1 &&
-        (prev as HTMLElement).contentEditable == "false" && (prev as HTMLElement).getBoundingClientRect().top >= coords.top)
-      offset--
+    if (browser.webkit && offset && node.nodeType == 1) {
+      const prev = node.childNodes[offset - 1];
+
+      if (prev?.nodeType == 1 && (prev as HTMLElement)?.contentEditable == "false" && (prev as HTMLElement)?.getBoundingClientRect().top >= coords.top) {
+      offset--;
+      }
+    }
     // Suspiciously specific kludge to work around caret*FromPoint
     // never returning a position at the end of the document
     if (node == view.dom && offset == node.childNodes.length - 1 && node.lastChild!.nodeType == 1 &&


marijnh:
The only way this seems to be possible is if `offset` is either negative or pointing beyond the length of the node. Both would suggest something is going wrong before this point, and I don't like adding tests like this to paper over other bugs. So I'd be interested in an example of a situation that triggers this crash.

nanov94:
Hi @marijnh :)
Thank you for your comment 🙏 

### **Summary**:
We are encountering an error (below) when selecting table cells, particularly when the table contains empty rows.
_Error: TypeError: Cannot read properties of undefined (reading 'nodeType')_

### **More details:**
Yes, you are correct about the case where `offset` is either negative or points beyond the length of the node. In my app, we have another case where `offset` is 0 and `node` has no children.

**Why is this the case? Empty rows without table cells:**  
_Firstly_, a `tr` element in HTML can have empty content [here](https://html.spec.whatwg.org/multipage/tables.html#the-tr-element). This mean that `node.childNodes` is empty array:
1. `node` does not have any `children` => `node.childNodes` = []
2. we got element by index in empty array => `node.childNodes[offset - 1]` = undefined
3. after got field from undefined element => `node.childNodes[offset - 1]).nodeType`

_Secondly_, this happened after using another prosemirror library: [prosemirror-tables](https://github.com/ProseMirror/prosemirror-tables).
**How can we create tables with empty rows via prosemirror?** You need to select a few table cells and rows and merge them. We use [mergeCells](https://github.com/ProseMirror/prosemirror-tables/blob/582b4e45b70da49472eed91698e5d3ecfbfcf5eb/src/commands.ts#L406) from `prosemirror-tables` for it (video attached)
[Demo merge table.webm](https://github.com/user-attachments/assets/e7e793ac-f25a-436f-8c68-2477ebbb935d)

marijnh:
> where offset is 0 and node has no children.

The line already contains `&& offset`, which will be false when `offset == 0`, so the code that could cause this crash will simply not execute in that situation.

nanov94:
> The line already contains && offset, which will be false when offset == 0, so the code that could cause this crash will simply not execute in that situation.

Yep, @marijnh, you are right, sorry for confusing 🙏 

We catch error when `offset` more than 0 and `node.childNodes` does not have children

image

marijnh:
What version of prosemirror-view are you using? In anything older than 1.31.0 there might be a situation where this happens, on Chrome when the document contains an `<input>` or `<textarea>` element.

nanov94:
we use "prosemirror-view": "1.33.5",

marijnh:
In that case, I'd really want to see instructions on how to reproduce this error, so that I can figure out what the underlying issue is.

PR #176, opened 2024-11-20
Support MarkView.ignoreMutation

ocavue:
This PR adds `MarkView.ignoreMutation()` method, similar to existing `NodeView.ignoreMutation()`. 

This is useful if users want to render and update the mark view dom element based on an extra state (like React's context) without triggering prosemirror's view update. 

---

This PR also includes some other minor updates:

- Add a new type `ViewMutationRecord`. This can reduce `as any` from ProseMirror's code and also users' code. 
- Add test for both `MarkView.ignoreMutation()` and `NodeView.ignoreMutation()`.
- Add a `flush(view)` API for easier test.

### DIFF
--- a/src/domobserver.ts
+++ b/src/domobserver.ts
@@ -151,8 +151,8 @@ export class DOMObserver {
     let desc = container && this.view.docView.nearestDesc(container)
     if (desc && desc.ignoreMutation({
       type: "selection",
-      target: container!.nodeType == 3 ? container!.parentNode : container
-    } as any)) {
+      target: container!.nodeType == 3 ? container!.parentNode! : container!
+    })) {
       this.setCurSelection()
       return true
     }
--- a/src/index.ts
+++ b/src/index.ts
@@ -14,7 +14,7 @@ import {DOMSelection, DOMNode, DOMSelectionRange, deepActiveElement, clearReused
 import * as browser from "./browser"
 
 export {Decoration, DecorationSet, DecorationAttrs, DecorationSource} from "./decoration"
-export {NodeView, MarkView} from "./viewdesc"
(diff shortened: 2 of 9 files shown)


ocavue:
Not related to this PR, but [this test case](https://github.com/ProseMirror/prosemirror-view/blob/5c33b883d39d7d40097e1fcdf17f543d18ad502a/test/webtest-selection.ts#L148) fails on my machine (macOS 15.1). It fails on both `master` branch and my development branch. 


```
  1) Browser tests (chrome)

         EditorView
           produces sensible screen coordinates in corner cases:
     Error: Uncaught Error: 12 != 13 at http://localhost:63203/_m/test/webtest-selection.js:150:21 (http://localhost:63203/_m/__/node_modules/ist/ist.js:42)
      at Context.<anonymous> (/Users/ocavue/code/github/prosemirror/node_modules/@marijn/testtool/src/testtool.js:50:32)
      at process.processImmediate (node:internal/timers:476:21)
```

src/domobserver.ts
```diff
@@ -151,8 +151,8 @@ export class DOMObserver {
     let desc = container && this.view.docView.nearestDesc(container)
     if (desc && desc.ignoreMutation({
       type: "selection",
-      target: container!.nodeType == 3 ? container!.parentNode : container
-    } as any)) {
+      target: container!.nodeType == 3 ? container!.parentNode! : container!
+    })) {
```
marijnh, line 155: This doesn't actually solve the type error. It also has nothing to do with this feature, so it's better to not lump it into the PR.

marijnh:
Looks good. Merged.

PR #178, opened 2025-02-14
Remove internal ViewDesc API from type declarations

ocavue:
In [prosemirror-view/dist/index.d.ts](https://unpkg.com/prosemirror-view@1.38.0/dist/index.d.ts), I noticed some internal APIs:

```ts
declare global {
    interface Node {
        pmViewDesc?: ViewDesc;
    }
}

declare class ViewDesc {
  // ...
}

declare class NodeViewDesc extends ViewDesc {
  // ...
}
```

While most of these APIs are fine, `pmViewDesc` is causing issues in my automated documentation generation from `.d.ts` files. To resolve this, I propose adding `/// @internal` comment to `pmViewDesc`. This will prevent them from appearing in the generated documentation.

### DIFF
--- a/src/viewdesc.ts
+++ b/src/viewdesc.ts
@@ -7,7 +7,10 @@ import {Decoration, DecorationSource, WidgetConstructor, WidgetType, NodeType} f
 import {EditorView} from "./index"
 
 declare global {
-  interface Node { pmViewDesc?: ViewDesc }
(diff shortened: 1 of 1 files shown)


marijnh:
Oh wow, I hadn't realized TS copied those into our .d.ts files. Thanks for spotting that.

PR #180, opened 2025-04-17
fix: add ariaHidden to br hack nodes

nperez0111:
In the same spirit as https://github.com/ProseMirror/prosemirror-view/pull/118 this makes the BR hack node be hidden to screen readers

### DIFF
--- a/src/viewdesc.ts
+++ b/src/viewdesc.ts
@@ -1383,7 +1383,10 @@ class ViewTreeUpdater {
         dom.className = "ProseMirror-separator"
         ;(dom as HTMLImageElement).alt = ""
       }
-      if (nodeName == "BR") dom.className = "ProseMirror-trailingBreak"
+      if (nodeName == "BR") {
+        dom.className = "ProseMirror-trailingBreak"
+        dom.ariaHidden = 'true'
+      }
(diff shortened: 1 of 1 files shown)


marijnh:
These nodes are present in most web-based editors, and screen readers seem to be able to handle them gracefully in my experience. What kind of issues are you running into, when using a screen reader, that would require this attribute?

miguelcaro:
@marijnh one of our clients report that the <br> it is been announced by the screen reader. 
please  take a look on the following image

another thing to solve this issue is add visibility: hidden for <br>

Screenshot 2025-04-24 at 8 23 43 AM

c.c. @nperez0111

marijnh:
How does the reader announce these? What do they sound like?

miguelcaro:
> How does the reader announce these? What do they sound like?

Don't know to be honest they just sent us this image with <Br> been announced by the NVDA speech viewer, like you can see in the image above

marijnh:
What I'm seeing with NVDA is that it says 'blank' on blank lines, whether they have the `<br>` or not, and nothing more. Until I actually understand the problem, I'm afraid I'm not going to merge this patch.

PR #182, opened 2025-05-14
When starting a composition, only delete the selection when selecting multiple lines

synw1024:
After reading this issue: https://github.com/ProseMirror/prosemirror/issues/1482. I'm sure the patch [4d2e1ba](https://github.com/ProseMirror/prosemirror-view/commit/4d2e1ba7545328313127a4ed94860b83e582a943) cause a bug to Sogou(https://shurufa.sogou.com) chinese input. It's partly the result of Sogou itself which doesn't sent "keydown" and "compositionupdate" events(https://github.com/electron/electron/issues/33386). And it's partly the result of the patch deletes the selection when starting a composition.

Bug reshow:
reproduce

Bug fixed:
fixed

### DIFF
--- a/src/input.ts
+++ b/src/input.ts
@@ -547,7 +547,7 @@ export function endComposition(view: EditorView, restarting = false) {
   if (restarting || view.docView && view.docView.dirty) {
     let sel = selectionFromDOM(view)
     if (sel && !sel.eq(view.state.selection)) view.dispatch(view.state.tr.setSelection(sel))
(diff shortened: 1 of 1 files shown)


marijnh:
Good point. I've pushed a slightly different patch (attached), since node object identity isn't something you can rely on, and inline nodes with content exist. Could you verify whether that approach works for you?

synw1024:
> Good point. I've pushed a slightly different patch (attached), since node object identity isn't something you can rely on, and inline nodes with content exist. Could you verify whether that approach works for you?

It works, I have tested

marijnh:
Great. I've tagged it as 1.39.3

PR #189, opened 2026-01-26
Fix hasFocus() for editors nested in contenteditable elements

excursus:
## Problem

When a ProseMirror editor is nested inside another element with `contenteditable="true"`, `view.hasFocus()` incorrectly returns `false` even when the editor has focus. This causes various functionality to break, including selection handling.

## Cause

When the editor is inside a contenteditable ancestor, the browser sets `document.activeElement` to the outermost contenteditable element rather than the editor's DOM element. The existing `hasFocus()` implementation compares `activeElement` directly to `this.dom`, which fails in this nested scenario.

## Solution

Added an `effectiveActiveElement()` helper method that accounts for nested contenteditable elements. When `activeElement` is an ancestor that contains `this.dom`, the helper returns `this.dom` as the effective active element. The `hasFocus()` method now uses this helper instead of directly accessing `this.root.activeElement`.

## Testing

- Added a new test case "reports focus correctly when nested in contenteditable" that verifies `hasFocus()` returns true when the editor is mounted inside a contenteditable wrapper
- Wrapped the entire test harness in a contenteditable element to validate behavior; several tests were failing before this fix, and now all pass.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -313,6 +313,20 @@ export class EditorView {
     }
   }
 
+  /// @internal
+  /// Returns the active element, accounting for nested contenteditable
+  /// elements. When the editor is inside another contenteditable,
+  /// activeElement may be the outer element even when the selection is
+  /// inside this editor.
+  effectiveActiveElement(): Element | null {
+    let active = this.root.activeElement
+    // If activeElement is an ancestor contenteditable that contains
+    // this.dom, treat this.dom as the effective active element.
+    if (active && active != this.dom && active.contains(this.dom))
+      return this.dom
+    return active
+  }
+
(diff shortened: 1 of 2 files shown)


marijnh:
I just verified that, on MacOS Chrome, focusing an editable element inside another editable element sets `document.activeElement` to the inner element. All other browsers behave the same. This is the test I used.

```
<div contenteditable=true id=outer>
Outer content
<div contenteditable=false>
  uneditable
  <div contenteditable=true id=inner>inner content</div>
</div>
</div>
```

Possibly you don't have an uneditable element around the inner editable element. That will make it part of the outer editable region.

PR #192, opened 2026-03-25
Check `nodeType` before calling `hasAttribute` in `NodeViewDesc.create`

### LINKED ISSUE #1566: `NodeViewDesc.create` crashes on Text DOM nodes when `toDOM` returns a string for non-text atom nodes
`NodeViewDesc.create` in `prosemirror-view` assumes that dom is an Element node when `node.isText` is false.

However, `DOMOutputSpec` explicitly allows string as a return type from `toDOM`, which produces a Text DOM node. When a non-text atom node's `toDOM` returns a string, ProseMirror crashes:

```
TypeError: dom.hasAttribute is not a function
```

This can be fixed with an extra check to make sure the node is an element.

```ts
dom.nodeType === Element.ELEMENT_NODE
```

https://github.com/ProseMirror/prosemirror-view/blob/1.41.6/src/viewdesc.ts#L708

### Repro

```ts
import { Schema } from "prosemirror-model";
import { EditorState } from "prosemirror-state";
import { EditorView } from "prosemirror-view";

const schema = new Schema({
  nodes: {
    doc: { content: "block+" },
    paragraph: {
      group: "block",
      content: "inline*",
      toDOM: () => ["p", 0],
      parseDOM: [{ tag: "p" }],
    },
    text: { group: "inline" },
    tag: {
      group: "inline",
      inline: true,
      atom: true,
      attrs: { value: { default: "" } },
      // Valid DOMOutputSpec — returns a string, which becomes a Text DOM node.
      // Crashes in NodeViewDesc.create because it calls dom.hasAttribute()
      // on the Text node.
      toDOM: (node) => node.attrs.value,
    },
  },
});

// Build a doc containing a tag node
const doc = schema.node("doc", null, [
  schema.node("paragraph", null, [
    schema.text("Hello "),
    schema.node("tag", { value: "{{ name }}" }),
    schema.text(", welcome!"),
  ]),
]);

const state = EditorState.create({ doc, schema });

// This crashes: TypeError: dom.hasAttribute is not a function
const view = new EditorView(document.querySelector("#editor"), { state });
```

[Codesandbox](https://codesandbox.io/p/sandbox/zcsjpg)

rohmann:
Fixes https://github.com/ProseMirror/prosemirror/issues/1566

### DIFF
--- a/src/viewdesc.ts
+++ b/src/viewdesc.ts
@@ -705,7 +705,7 @@ export class NodeViewDesc extends ViewDesc {
       let spec = (DOMSerializer.renderSpec as any)(document, node.type.spec.toDOM!(node), null, node.attrs)
       ;({dom, contentDOM} = spec as {dom: DOMNode, contentDOM?: HTMLElement})
     }
-    if (!contentDOM && !node.isText && dom.nodeName != "BR") { // Chrome gets confused by <br contenteditable=false>
+    if (!contentDOM && !node.isText && dom.nodeType == 3 && dom.nodeName != "BR") { // Chrome gets confused by <br contenteditable=false>
       if (!(dom as HTMLElement).hasAttribute("contenteditable")) (dom as HTMLElement).contentEditable = "false"
       if (node.type.spec.draggable) (dom as HTMLElement).draggable = true
     }


marijnh:
See https://github.com/ProseMirror/prosemirror-model/commit/042fe20a74e7d4dd9b664e9c02040fd411aedf31

